PR #13, opened 2022-03-02
Add extensions methods on string to serialize and deserialize objects

maxime-poulain:
Having extension method on strings to serialize and deserialize  object can be useful especially at the application layer.
Implementing them in the domain layer will make them available across the whole project.


Victor-van-Duijnen (approved):
Not entirely sure it will be the best layer to put that helper class. Are we sure it will only be used in the Domain layer of the application? Well it's nitpicking (though there is a deeper reflexion behind) and I know domain layer "purity" is a controversial topic so I will trust you on that one.

maxime-poulain:
You can find entreprise and Domain service in that layer, isn't  ? As soon as you have a bit of logic this can be handy in multiple situation.

PR #19, opened 2022-03-07
Add behaviour pipeline for logging in handlers

Victor-van-Duijnen:
- Add a generic logging behaviour for query and command handlers.
- Divide commands/queries for fetching and creating user inside cfa
-  Add integrations tests for user fetching/creation

### DIFF
--- a/Smart.FA.Catalog.IntegrationTests/Smart.FA.Catalog.IntegrationTests.csproj
+++ b/Smart.FA.Catalog.IntegrationTests/Smart.FA.Catalog.IntegrationTests.csproj
@@ -12,6 +12,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore" Version="6.0.1" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="6.0.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
+    <PackageReference Include="Moq" Version="4.16.1" />
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
@@ -25,6 +26,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Smart.FA.Catalog.Tests.Common\Smart.FA.Catalog.Tests.Common.csproj" />
+    <ProjectReference Include="..\src\Smart.FA.Catalog.Application\Application.csproj" />
     <ProjectReference Include="..\src\Smart.FA.Catalog.Infrastructure\Infrastructure.csproj" />
   </ItemGroup>
 
--- a/Smart.FA.Catalog.IntegrationTests/TrainingContext/Trainer_Tests.cs
+++ b/Smart.FA.Catalog.IntegrationTests/TrainingContext/Trainer_Tests.cs
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using AutoFixture;
 using Core.Domain;
 using FluentAssertions;
 using Microsoft.EntityFrameworkCore;
@@ -14,6 +15,7 @@ namespace Smart.FA.Catalog.IntegrationTests.TrainingContext;
 public class TrainerTests : IntegrationTestBase
 {
     private readonly TrainerFactory _trainerFactory = new();
+    private readonly UserFactory _userFactory = new();
 
     [Theory]
     [InlineData("Victor", "vD")]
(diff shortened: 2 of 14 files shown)


Smart.FA.Catalog.IntegrationTests/Smart.FA.Catalog.IntegrationTests.csproj
```diff
@@ -12,6 +12,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore" Version="6.0.1" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="6.0.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
+    <PackageReference Include="Moq" Version="4.16.1" />
```
LievenPraet, line 15: Can we use `NSubstitute` instead of `Moq`?
We want to stay consistent with the packages we use throughout projects.

Smart.FA.Catalog.IntegrationTests/TrainingContext/Trainer_Tests.cs
```diff
@@ -49,7 +51,19 @@ public async Task CanChangeTrainerName(string firstName, string lastName)
...
+        var trainer = _trainerFactory.CreateFromUser(user);
 
+        context.Trainers.Add(trainer);
+        await context.SaveChangesAsync();
 
-
+        trainer.Should().NotBeNull();
+        trainer!.Name.FirstName.Should().NotBeEmpty();
+        trainer.Name.LastName.Should().NotBeEmpty();
+        trainer.DefaultLanguage.Value.Should().NotBeEmpty();
```
LievenPraet, line 67: Instead of checking for not empty, you can check the actual values since you have them for your user object

Smart.FA.Catalog.IntegrationTests/TrainingContext/Trainer_Tests.cs
```diff
@@ -14,6 +15,7 @@ namespace Smart.FA.Catalog.IntegrationTests.TrainingContext;
 public class TrainerTests : IntegrationTestBase
 {
     private readonly TrainerFactory _trainerFactory = new();
+    private readonly UserFactory _userFactory = new();
```
LievenPraet, line 18: You can use `NSubstitute` to create these instances for you.
This will help in the future in case extra dependencies are introduced in these classes because `NSubstitute` will handle them for you.

Smart.FA.Catalog.IntegrationTests/TrainingContext/Trainer_Tests.cs
```diff
@@ -14,6 +15,7 @@ namespace Smart.FA.Catalog.IntegrationTests.TrainingContext;
...
 
     [Theory]
     [InlineData("Victor", "vD")]
```
LievenPraet, line 21: You can add additional tests to test what happens when you pass in empty or invalid values.

Smart.FA.Catalog.IntegrationTests/TrainingContext/Trainer_Tests.cs
```diff
@@ -49,7 +51,19 @@ public async Task CanChangeTrainerName(string firstName, string lastName)
...
+        await using var context = GivenTrainingContext();
+        var user = _userFactory.CreateClean();
+        var trainer = _trainerFactory.CreateFromUser(user);
 
+        context.Trainers.Add(trainer);
+        await context.SaveChangesAsync();
 
-
+        trainer.Should().NotBeNull();
+        trainer!.Name.FirstName.Should().NotBeEmpty();
```
LievenPraet, line 65: Any reason you use `!` for some lines but not other lines?

Smart.FA.Catalog.Tests.Common/UserFactory.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Core.Domain.Dto;
+using Core.Domain.Enumerations;
+
+namespace Smart.FA.Catalog.Tests.Common;
+
+public class UserFactory
+{
+    public UserDto CreateClean()
+    {
+        return new UserDto("1", "John", "Doe", ApplicationType.Account.Name);
```
LievenPraet, line 10: You could use `Fixture` here as well

LievenPraet (changes requested):

Smart.FA.Catalog.IntegrationTests/Smart.FA.Catalog.IntegrationTests.csproj
Victor-van-Duijnen: Alright I can do that, I will have to substitute it in the unit test projects, too so I will do a separate PR for it

Smart.FA.Catalog.IntegrationTests/TrainingContext/Trainer_Tests.cs
Victor-van-Duijnen: If the trainer is null, it should not pass line #64  in the assert section of the code.  So at line #65 I am absolutely certain the trainer is not null (before it, I can't be certain), so I just make it explicit to the compiler since the IDE would complain that if the trainer is null, he would not be able to access any of its properties.

maxime-poulain:
Please don't create a PR with three different features.
This should have been spitted into three PR.
Or at least don't give it 
>Add behaviour pipeline for logging in handlers
as title.

PR #24, opened 2022-03-09
Add behaviour pipeline for logging in handlers

Victor-van-Duijnen:
Generic logging behaviour for query and command handlers

### DIFF
--- a/src/Smart.FA.Catalog.Web/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Smart.FA.Catalog.Web/Extensions/ServiceCollectionExtensions.cs
@@ -1,4 +1,7 @@
+using Application.Interceptors;
 using Core.Services;
+using MediatR;
+using Web.Extensions.Middlewares;
 using Web.Identity;
 using Web.Options;
 
@@ -8,12 +11,13 @@ public static class ServiceCollectionExtensions
 {
     public static void AddApi(this IServiceCollection services, IConfigurationSection adminOptionsSection)
     {
-        services.AddScoped<IUserIdentity, UserIdentity>();
-        services.AddOptions(adminOptionsSection);
+        services
+            .AddScoped<IUserIdentity, UserIdentity>()
+            .AddOptions(adminOptionsSection);
     }
 
-    public static void AddOptions(this IServiceCollection services, IConfigurationSection adminOptionsSection)
+    private static IServiceCollection AddOptions(this IServiceCollection services, IConfigurationSection adminOptionsSection)
     {
-        services.Configure<AdminOptions>(adminOptionsSection);
+        return services.Configure<AdminOptions>(adminOptionsSection);
     }
 }
--- a/src/Smart.FA.Catalog.Application/Application.csproj
+++ b/src/Smart.FA.Catalog.Application/Application.csproj
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Folder Include="Interceptors" />
     <Folder Include="UseCases\Queries\" />
   </ItemGroup>
 
--- a/src/Smart.FA.Catalog.Application/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Smart.FA.Catalog.Application/Extensions/ServiceCollectionExtensions.cs
@@ -1,4 +1,5 @@
 using System.Reflection;
+using Application.Interceptors;
 using Application.SeedWork;
 using MediatR;
 using Microsoft.Extensions.DependencyInjection;
@@ -9,6 +10,13 @@ public static class ServiceCollectionExtensions
 {
     public static void AddApplication(this IServiceCollection services)
     {
-        services.AddMediatR(Assembly.GetExecutingAssembly());
+        services
+            .AddMediatR(Assembly.GetExecutingAssembly())
+            .AddPipelineBehaviours();
+    }
+
+    private static IServiceCollection AddPipelineBehaviours(this IServiceCollection services)
+    {
+        return services.AddTransient(typeof(IPipelineBehavior<,>), typeof(LoggingPipelineBehavior<,>));
     }
 }
--- a/src/Smart.FA.Catalog.Application/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Smart.FA.Catalog.Application/Extensions/ServiceCollectionExtensions.cs
@@ -12,10 +12,10 @@ public static class ServiceCollectionExtensions
     {
         services
             .AddMediatR(Assembly.GetExecutingAssembly())
-            .AddPipelineBehaviours();
+            .AddMediatrPipelineBehaviours();
     }
 
-    private static IServiceCollection AddPipelineBehaviours(this IServiceCollection services)
+    private static IServiceCollection AddMediatrPipelineBehaviours(this IServiceCollection services)
     {
         return services.AddTransient(typeof(IPipelineBehavior<,>), typeof(LoggingPipelineBehavior<,>));
     }
--- a/src/Smart.FA.Catalog.Application/Interceptors/LoggingPipelineBehavior.cs
+++ b/src/Smart.FA.Catalog.Application/Interceptors/LoggingPipelineBehavior.cs
@@ -1,0 +1,53 @@
+﻿using Application.SeedWork;
+using Core.Exceptions;
+using Core.LogEvents;
+using Core.SeedWork;
+using MediatR;
+using Microsoft.Extensions.Logging;
+
+namespace Application.Interceptors;
+
+/// <summary>
+/// Global logger applied on all handlers.
+/// It will format potential exception with the Error template when pertinent and set the response to a successful state, otherwise.
+/// </summary>
+/// <typeparam name="TRequest">The input object of the handler, it needs to inherit from IMediatr IRequest class</typeparam>
+/// <typeparam name="TResponse">the output object of the handler, it needs to inherit from ResponseBase and be instantiable</typeparam>
+public class LoggingPipelineBehavior<TRequest, TResponse> : IPipelineBehavior<TRequest, TResponse>
+    where TRequest : IRequest<TResponse> where TResponse : ResponseBase, new()
+{
+    private readonly ILogger<LoggingPipelineBehavior<TRequest, TResponse>> _logger;
+    public LoggingPipelineBehavior
+    (
+        ILogger<LoggingPipelineBehavior<TRequest, TResponse>> logger
+    )
+    {
+        _logger = logger;
+    }
+
+    public async Task<TResponse> Handle
+    (
+        TRequest request
+        , CancellationToken cancellationToken,
+        RequestHandlerDelegate<TResponse> next
+    )
+    {
+        TResponse response = new();
+        try
+        {
+            response = await next();
+        }
+        catch (DomainException exception)
+        {
+            response.AddError(exception.Error);
+            _logger.LogError(LogEventIds.ErrorEventId, exception.Error.Serialize());
+        }
+        catch (Exception exception)
+        {
+            response.AddError(Errors.General.UnexpectedHandlerError());
+            _logger.LogError(LogEventIds.ErrorEventId, exception.ToString());
+        }
+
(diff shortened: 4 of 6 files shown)


src/Smart.FA.Catalog.Web/Extensions/Middlewares/HandlerLoggerPipelineBehavior.cs
```diff
@@ -0,0 +1,54 @@
+﻿using Application.SeedWork;
+using Core.Exceptions;
+using Core.LogEvents;
+using Core.SeedWork;
+using MediatR;
+
+namespace Web.Extensions.Middlewares;
```
LievenPraet, line 7: Why not include Smart FA Catalog in the namespace?

maxime-poulain: I would not put the class at the level of the presentation layer but closely to the application one.

This is more a application concern in the end than a web one.

Also if kept in the web projet I would definitely not put that class inside a Extensions/**/** folder.
The class isn't extending some class methods.

This is not a actual middleware neither it is an interceptor at the Mediatr(' lib.

src/Smart.FA.Catalog.Web/Extensions/Middlewares/HandlerLoggerPipelineBehavior.cs
```diff
@@ -0,0 +1,54 @@
...
+
+/// <summary>
+/// Global logger applied on all handlers.
+/// It will format potential exception with the Error template when pertinent and set the response to a successful state, otherwise.
+/// </summary>
+/// <typeparam name="TRequest">The input object of the handler, it needs to inherit from IMediatr IRequest class</typeparam>
+/// <typeparam name="TResponse">the output object of the handler, it needs to inherit from ResponseBase and be instantiable</typeparam>
+public class HandlerLoggerPipelineBehavior<TRequest, TResponse> : IPipelineBehavior<TRequest, TResponse>
+    where TRequest : IRequest<TResponse> where TResponse : ResponseBase, new()
```
maxime-poulain, line 16: It's a pipeline behavior as stated by the interface, it is not an actual handler.
At least something that handles another thing when a particular condition happens.

I would remove the Prefix "Handler"

src/Smart.FA.Catalog.Web/Extensions/Middlewares/HandlerLoggerPipelineBehavior.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    public HandlerLoggerPipelineBehavior
+    (
+        ILogger<HandlerLoggerPipelineBehavior<TRequest, TResponse>> logger
+    )
+    {
+        _logger = logger;
+    }
+
+    public async Task<TResponse> Handle
```
maxime-poulain, line 28: The formatting is , at least to me, hard to read
```suggestion
    public async Task<TResponse> Handle(TRequest request, CancellationToken cancellationToken, RequestHandlerDelegate<TResponse> next
    )
```
or
```suggestion
    public async Task<TResponse> Handle(IRequest request,
        CancellationToen token,
        RequesthandlerDelegate next)
    
```
would be more suitable

src/Smart.FA.Catalog.Web/Extensions/Middlewares/HandlerLoggerPipelineBehavior.cs
```diff
@@ -0,0 +1,54 @@
...
+        , CancellationToken cancellationToken,
+        RequestHandlerDelegate<TResponse> next
+    )
+    {
+        TResponse response = new();
+        try
+        {
+            response = await next();
+            response.SetSuccess();
+        }
```
maxime-poulain, line 40: The concern, or at least what I expect from this pipeline behavior, is to do some kind of logging.

I should not be its responsibility to determine if the state of an operation is successful or not.

The underlying actual command/query handler should be responsible to set that thing up.

src/Smart.FA.Catalog.Web/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -8,12 +10,21 @@ public static class ServiceCollectionExtensions
...
+
         services.AddScoped<IUserIdentity, UserIdentity>();
         services.AddOptions(adminOptionsSection);
+        services.AddCustomMediatR();
     }
 
-    public static void AddOptions(this IServiceCollection services, IConfigurationSection adminOptionsSection)
+    public static IServiceCollection AddOptions(this IServiceCollection services, IConfigurationSection adminOptionsSection)
     {
         services.Configure<AdminOptions>(adminOptionsSection);
```
maxime-poulain, line 21: You can directly return this line.

src/Smart.FA.Catalog.Web/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -8,12 +10,21 @@ public static class ServiceCollectionExtensions
 {
     public static void AddApi(this IServiceCollection services, IConfigurationSection adminOptionsSection)
     {
+
         services.AddScoped<IUserIdentity, UserIdentity>();
         services.AddOptions(adminOptionsSection);
+        services.AddCustomMediatR();
     }
```
maxime-poulain, line 17: I would replace the void modifier to IServiceCollection and chain the factory pattern with a return

src/Smart.FA.Catalog.Web/Extensions/Middlewares/HandlerLoggerPipelineBehavior.cs
```diff
@@ -0,0 +1,54 @@
...
+            _logger.LogError(LogEventIds.ErrorEventId, exception.Error.Serialize());
+        }
+        catch (Exception exception)
+        {
+            response.AddError(Errors.General.UnexpectedHandlerError());
+            _logger.LogError(LogEventIds.ErrorEventId, exception.Message);
+        }
+
+        return response;
+    }
```
maxime-poulain, line 53: I don't think capturing globally any unexpected exception is a good practice.
we should let the web / presentation layer to be responsible how to handle "500" error. (such as redirect or display some kind of generic message to the end user).

src/Smart.FA.Catalog.Web/Extensions/Middlewares/HandlerLoggerPipelineBehavior.cs
```diff
@@ -0,0 +1,54 @@
...
+        catch (DomainException exception)
+        {
+            response.AddError(exception.Error);
+            _logger.LogError(LogEventIds.ErrorEventId, exception.Error.Serialize());
+        }
+        catch (Exception exception)
+        {
+            response.AddError(Errors.General.UnexpectedHandlerError());
+            _logger.LogError(LogEventIds.ErrorEventId, exception.Message);
+        }
```
maxime-poulain, line 50: You should log the exception itself, otherwise by just logging the exception's message you loose important information such as the stacktrace.

src/Smart.FA.Catalog.Web/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -8,12 +10,21 @@ public static class ServiceCollectionExtensions
...
 
-    public static void AddOptions(this IServiceCollection services, IConfigurationSection adminOptionsSection)
+    public static IServiceCollection AddOptions(this IServiceCollection services, IConfigurationSection adminOptionsSection)
     {
         services.Configure<AdminOptions>(adminOptionsSection);
+        return services;
+    }
+
+    public static IServiceCollection AddCustomMediatR(this IServiceCollection services)
+    {
```
maxime-poulain, line 26: Hm, I thin there is something wrong with that name.
We are not actually adding to the IoC container a custom Mediatr (i.e. custom implementation of the MediatR library) but  some other thing like pipeline, etc.

I would definitely call it "AddMediatrPipleBehaviors" or siomething.
@LievenPraet 
You might come up with another suggestion

maxime-poulain, line 26: I thin it would be more appropriate to have this method inside this class:
https://github.com/smartcoop/smart-cfa/blob/feature/develop/src/Smart.FA.Catalog.Application/Extensions/ServiceCollectionExtensions.cs

src/Smart.FA.Catalog.Web/Extensions/Middlewares/HandlerLoggerPipelineBehavior.cs
Victor-van-Duijnen: Ok, it's auto-formatted by Jetbrain Rider, I will have to do something about my IDE settings

src/Smart.FA.Catalog.Web/Extensions/Middlewares/HandlerLoggerPipelineBehavior.cs
Victor-van-Duijnen: Good point, I will remove SetSuccess from there.

src/Smart.FA.Catalog.Web/Extensions/Middlewares/HandlerLoggerPipelineBehavior.cs
Victor-van-Duijnen: Hmm the presentation layer could and should still handle it. The logger is catching the exception but it's adding the error to the response object (which in turn will be indicated in a state of error). How the presentation layer handle the error is its responsibility but at least, the error message will be formatted on our terms with a log event id.

src/Smart.FA.Catalog.Web/Extensions/ServiceCollectionExtensions.cs
Victor-van-Duijnen: Yes actually I was thinking more along the lines of "AddPipelineBehaviour" (the fact that we chose Mediatr to handle that is irrelevant)

maxime-poulain: Hm it would be relevant to have the MediatR name in the method that register stuff in the IoC container.
The reason is that the pipeline behavior is something that belong to the library itself.
We might tomorrow switch to Super "Super Mediator" , this library might not propose the concept of pipelines.

Victor-van-Duijnen: I am not sure I understand your reasoning. It's true behaviour pipeline is a mediatr concept so it would kinda redundant to put it in the name in my opinion. If we were using "Super Mediator" instead of Mediatr and it doesn't support the concept of behaviour pipelines you would have to rename or remove the extension method in both case, anyway.
Now I am even thinking about calling it something like AddInterceptors or something that would make sense without having to know the library or the made-up name the library coined to describe an existing thing.

maxime-poulain: In my opinion, when it comes to a technology that is not part of the framework, a good practices is to do some something like this:
image.

An even better thing to do is even to have a dedicated project to will handle registration for the IoC container not to have any projects / layers to be responsible to do that.

Victor-van-Duijnen: Ok I have renamed the method to move along the PR, then

src/Smart.FA.Catalog.Web/Extensions/Middlewares/HandlerLoggerPipelineBehavior.cs
Victor-van-Duijnen: I have moved the class into the application layer.
@LievenPraet can we schedule a small meeting to discuss namespaces, I can make the change to another guidelines for namespaces in my projects (in another PR because it will affect all files) but I want to discuss it before with you!

Victor-van-Duijnen:
Same comment as previous PR, I have taken account of all comments so I will go ahead and merge the changes.

PR #26, opened 2022-03-09
Divide proxy header middleware in multiple commands/queries

Victor-van-Duijnen:
Separate fetching an user and its potential linked trainer from the user id and from creating the trainer if no trainer is linked to an userId/applicationtypeid in order to have a good command/query division (queries should not update data)

### DIFF
--- a/src/Smart.FA.Catalog.Application/UseCases/Commands/CreateTrainerFromUserAppCommandHandler.cs
+++ b/src/Smart.FA.Catalog.Application/UseCases/Commands/CreateTrainerFromUserAppCommandHandler.cs
@@ -1,0 +1,74 @@
+﻿using Application.SeedWork;
+using Core.Domain;
+using Core.Domain.Dto;
+using Core.Domain.Enumerations;
+using Core.LogEvents;
+using Core.SeedWork;
+using Infrastructure.Persistence;
+using MediatR;
+using Microsoft.Extensions.Logging;
+
+namespace Application.UseCases.Commands;
+
+public class
+    CreateTrainerFromUserAppCommandHandler : IRequestHandler<CreateTrainerFromUserAppRequest,
+        CreateTrainerFromUserAppResponse>
+{
+    private readonly ILogger<CreateTrainerFromUserAppCommandHandler> _logger;
+    private readonly CatalogContext _context;
+
+    public CreateTrainerFromUserAppCommandHandler
+    (
+        ILogger<CreateTrainerFromUserAppCommandHandler> logger
+        , CatalogContext context
+    )
+    {
+        _logger = logger;
+        _context = context;
+    }
+
+    public async Task<CreateTrainerFromUserAppResponse> Handle(CreateTrainerFromUserAppRequest command,
+        CancellationToken cancellationToken)
+    {
+        CreateTrainerFromUserAppResponse appResponse = new();
+
+        var linkedTrainer = new Trainer
+        (
+            Name.Create
+            (
+                command.User.FirstName
+                , command.User.LastName
+            ).Value
+            , TrainerIdentity.Create
+            (
+                command.User.UserId
+                , Enumeration.FromDisplayName<ApplicationType>(command.User.ApplicationType)
+            ).Value
+            , string.Empty
+            , string.Empty
+            , Language.Create("EN").Value
+        );
+        _context.Trainers.Add(linkedTrainer);
+        await _context.SaveChangesAsync(cancellationToken);
+
+        _logger.LogInformation(LogEventIds.TrainerCreated,
+            "A new trainer for user id {UserId}, application type {ApplicationType} was created", command.User.UserId,
+            command.User.ApplicationType);
+
+        appResponse.Trainer = linkedTrainer;
+
+        appResponse.SetSuccess();
+
+        return appResponse;
+    }
+}
+
+public class CreateTrainerFromUserAppRequest : IRequest<CreateTrainerFromUserAppResponse>
+{
+    public UserDto User { get; set; } = null!;
+}
+
+public class CreateTrainerFromUserAppResponse : ResponseBase
+{
+    public Trainer Trainer { get; set; } = null!;
+}
--- a/src/Smart.FA.Catalog.Application/UseCases/Queries/GetTrainerFromUserAppQueryHandler.cs
+++ b/src/Smart.FA.Catalog.Application/UseCases/Queries/GetTrainerFromUserAppQueryHandler.cs
@@ -1,8 +1,8 @@
 using Application.SeedWork;
 using Core.Domain;
+using Core.Domain.Dto;
 using Core.Domain.Enumerations;
-using Core.SeedWork;
-using Core.Services;
+using Core.Exceptions;
 using Infrastructure.Persistence;
 using Infrastructure.Services;
 using MediatR;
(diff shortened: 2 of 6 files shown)


src/Smart.FA.Catalog.Core/Domain/Training/Training.cs
```diff
@@ -40,15 +40,15 @@ public class Training : Entity, IAggregateRoot
...
         , IEnumerable<TrainingType> types
         , IEnumerable<TrainingSlotNumberType> slotNumberTypes
         , IEnumerable<TrainingTargetAudience> targetAudiences
         , IEnumerable<TrainingTopic> topics
     )
     {
-        AddDetails(trainingDetail.Title!, trainingDetail.Goal!, trainingDetail.Methodology!,
-            Language.Create(trainingDetail.Language).Value);
+        AddDetails(detail.Title!, detail.Goal!, detail.Methodology!,
+            Language.Create(detail.Language).Value);
```
LievenPraet, line 51: When you access `Value`, are you confident that `Language.Create` will never return null?

src/Smart.FA.Catalog.Core/Domain/Training/Training.cs
```diff
@@ -40,15 +40,15 @@ public class Training : Entity, IAggregateRoot
...
+        , TrainingDetailDto detail
         , IEnumerable<TrainingType> types
         , IEnumerable<TrainingSlotNumberType> slotNumberTypes
         , IEnumerable<TrainingTargetAudience> targetAudiences
         , IEnumerable<TrainingTopic> topics
     )
     {
-        AddDetails(trainingDetail.Title!, trainingDetail.Goal!, trainingDetail.Methodology!,
-            Language.Create(trainingDetail.Language).Value);
+        AddDetails(detail.Title!, detail.Goal!, detail.Methodology!,
```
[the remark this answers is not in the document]
maxime-poulain: You can't.
However the AddDetails method maes a check with a Guard clause on Title(doesn"t ensure that the goal or methodoligy variable are null )

src/Smart.FA.Catalog.Core/Domain/Training/Training.cs
```diff
@@ -40,15 +40,15 @@ public class Training : Entity, IAggregateRoot
     public Training
```
maxime-poulain, line 40: Like stated in another PR,
this formatting is quite hard to read, at least, to my opinion.

src/Smart.FA.Catalog.Application/UseCases/Queries/GetTrainerFromUserAppQueryHandler.cs
```diff
@@ -1,67 +1,46 @@
...
-            var userStrategy =  _userStrategyResolver.Resolve(request.ApplicationType!);
-            var user = await userStrategy.GetAsync(request.UserId!);
-            var linkedTrainer =
-                await _catalogContext.Trainers.FirstOrDefaultAsync(trainer => trainer.Identity.UserId == request.UserId, cancellationToken);
-            if (linkedTrainer is null)
-            {
-                linkedTrainer = new Trainer(Name.Create(user.FirstName, user.LastName).Value,
-                    TrainerIdentity.Create(user.UserId, Enumeration.FromDisplayName<ApplicationType>(user.ApplicationType)).Value,string.Empty, string.Empty,
-                    Language.Create("EN").Value);
-                await _catalogContext.Trainers.AddAsync(linkedTrainer, cancellationToken);
```
maxime-poulain, line 42: I would add to the PR an explanation of what it is all about.
Like stated the person who theorize CQS:
"Asking a question should not change the answer" 
A query purpose is to retrieve some result but not change the state of your application.

Therefore that's why this PR has been introduced.

src/Smart.FA.Catalog.Application/UseCases/Queries/GetUserAppFromIdQueryHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        var userStrategy = _userStrategyResolver.Resolve(query.ApplicationType!);
+        resp.User = await userStrategy.GetAsync(query.UserId!);
+        resp.SetSuccess();
+
+        return resp;
+    }
+}
+
+public class GetUserAppFromIdRequest : IRequest<GetUserAppFromIdResponse>
```
maxime-poulain, line 40: The suffix of this class should not be "Request" but rather Query.
The file is named "GetUserAppFromIdQueryHandler"
I expected to have something that handles a "GetUserAppFromIdQuery" class.

Furthermore it makes more sense to name objects sent by Mediatr with Command/Query suffix than Request.
I know the API of Mediator uses IRequest, IRequestHandler.

But the naming of chosen pattern should precede the API naming of a library used to implement that pattern.

src/Smart.FA.Catalog.Application/UseCases/Queries/GetUserAppFromIdQueryHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+using Core.Domain.Enumerations;
+using Core.Exceptions;
+using Core.Extensions;
+using Infrastructure.Services;
+using MediatR;
+using Microsoft.Extensions.Logging;
+
+namespace Application.UseCases.Queries;
+
+public class GetUserAppFromIdQueryHandler : IRequestHandler<GetUserAppFromIdRequest, GetUserAppFromIdResponse>
```
maxime-poulain, line 12: I think the filename should not be 
>GetUserAppFromIdQueryHandler 
but something like
>GetUserAppFromIdQuery

An handler by definition is "an abstraction" , something that you won't use / manipulate in the code.

What  you will manipulate however is a Query / Command / IRequest.
To me, it maes more sense, if the Command/Query are in the same file than the Handler to be named by the query / command object

src/Smart.FA.Catalog.Web/Extensions/Middlewares/ProxyHeaderMiddleware.cs
```diff
@@ -19,15 +25,21 @@ public ProxyHeaderMiddleware(RequestDelegate next)
...
-
         var userId = (string.IsNullOrEmpty(context.Request.Headers["user_id"].ToString())  ? "1" : context.Request.Headers["user_id"].ToString())!;
         var appName =  string.IsNullOrEmpty(context.Request.Headers["app_name"].ToString()) ? ApplicationType.Default.Name : context.Request.Headers["app_name"].ToString();
         var applicationType = Enumeration.FromDisplayName<ApplicationType>(appName);
 
-        var resp = await mediator.Send(new GetTrainerFromUserAppRequest {UserId = userId, ApplicationType = applicationType});
+        var userResponse = await mediator.Send(new GetUserAppFromIdRequest{UserId = userId, ApplicationType = applicationType});
+        var trainerResponse = await mediator.Send(new GetTrainerFromUserAppRequest {User = userResponse.User});
+
+        if (trainerResponse.Trainer is null)
```
maxime-poulain, line 35: Could it be possible to refactor the above lines into one single query ?

src/Smart.FA.Catalog.Application/UseCases/Commands/CreateTrainerFromUserAppCommandHandler.cs
```diff
@@ -0,0 +1,68 @@
...
+    )
+    {
+        _logger = logger;
+        _context = context;
+    }
+
+    public async Task<CreateTrainerFromUserAppResponse> Handle(CreateTrainerFromUserAppRequest command,
+        CancellationToken cancellationToken)
+    {
+        CreateTrainerFromUserAppResponse appResponse = new();
```
maxime-poulain, line 32: You imight want to log a new record in the database will be created because it is the first time he/she logs.

src/Smart.FA.Catalog.Core/Domain/Training/Training.cs
Victor-van-Duijnen: Yes that's an absolute certitude Value won't be null. It's in the way the ValueObject work, if it was not the case, the object would not be created at all and it would throw an exception earlier.

src/Smart.FA.Catalog.Core/Domain/Training/Training.cs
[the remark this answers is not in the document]
Victor-van-Duijnen: Good point, it's a relica from before a training could not be in a draft state (my PR only was about the renaming of the var but since I am at it, I will do that here too). Now all you need is a title in order to create a training in draft (that's why the Guard clause only applies on Title) but it won't go in a published state unless all required field (like goal and methodology are filled). I will change the Training class to reflect that!

src/Smart.FA.Catalog.Web/Extensions/Middlewares/ProxyHeaderMiddleware.cs
```diff
@@ -19,15 +25,21 @@ public ProxyHeaderMiddleware(RequestDelegate next)
 
     public async Task InvokeAsync(HttpContext context, IMediator mediator )
     {
-
         var userId = (string.IsNullOrEmpty(context.Request.Headers["user_id"].ToString())  ? "1" : context.Request.Headers["user_id"].ToString())!;
```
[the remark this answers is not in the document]
Victor-van-Duijnen: That was just because I needed to have something published before the all Nginx system was actually up and running. In the real-world scenario, there would not be default value. I will comment!

src/Smart.FA.Catalog.Application/UseCases/Queries/GetTrainerFromUserAppQueryHandler.cs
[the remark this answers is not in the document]
Victor-van-Duijnen: Good to know! I will change that right away.

Victor-van-Duijnen:
I will be merging that, I have taken into account every one of your comment (except for the change whose scope is too big for that PR that I will introduce in following PR)

PR #27, opened 2022-03-09
Rector factory method and mocking of test project

Victor-van-Duijnen:
- Make factory test methods static
- Replace Moq by NSubstitute

### DIFF
--- a/Smart.FA.Catalog.IntegrationTests/TrainingContext/Trainer_Tests.cs
+++ b/Smart.FA.Catalog.IntegrationTests/TrainingContext/Trainer_Tests.cs
@@ -1,9 +1,11 @@
-using System.Collections.Generic;
-using System.Linq;
+using System;
 using System.Threading.Tasks;
+using AutoFixture;
 using Core.Domain;
+using Core.Domain.Dto;
+using Core.Domain.Enumerations;
+using Core.SeedWork;
 using FluentAssertions;
-using Microsoft.EntityFrameworkCore;
 using Smart.FA.Catalog.IntegrationTests.Base;
 using Smart.FA.Catalog.Tests.Common;
 using Xunit;
@@ -13,14 +15,14 @@ namespace Smart.FA.Catalog.IntegrationTests.TrainingContext;
 [Collection("Integration test collection")]
 public class TrainerTests : IntegrationTestBase
 {
-    private readonly TrainerFactory _trainerFactory = new();
+    private readonly Fixture _fixture = new();
 
     [Theory]
     [InlineData("Victor", "vD")]
     public async Task CanCreateTrainer(string firstName, string lastName)
     {
         await using var context = GivenTrainingContext();
-        var trainer = _trainerFactory.Create(firstName, lastName);
+        var trainer = TrainerFactory.Create(firstName, lastName);
 
         context.Trainers.Attach(trainer);
         await context.SaveChangesAsync();
@@ -38,7 +40,7 @@ public class TrainerTests : IntegrationTestBase
     public async Task CanChangeTrainerName(string firstName, string lastName)
     {
         await using var context = GivenTrainingContext();
-        var trainer = _trainerFactory.Create(firstName, lastName);
+        var trainer = TrainerFactory.Create(firstName, lastName);
         context.Trainers.Attach(trainer);
         await context.SaveChangesAsync();
         var newName = Name.Create("Maxime", "P.");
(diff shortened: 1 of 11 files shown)


Smart.FA.Catalog.IntegrationTests/TrainingContext/Trainer_Tests.cs
```diff
@@ -49,7 +51,39 @@ public async Task CanChangeTrainerName(string firstName, string lastName)
...
+        trackedTrainer.Should().NotBeNull();
+        trackedTrainer!.Name.Should().BeEquivalentTo(trainer.Name);
+        trackedTrainer.DefaultLanguage.Should().BeEquivalentTo(trainer.DefaultLanguage);
+    }
 
+    [Theory]
+    [InlineData(null, null)]
+    [InlineData("1", null)]
+    [InlineData(null, 1)]
+    public async Task CantCreateFromInvalidUser(string? userId, int? applicationTypeId)
```
LievenPraet, line 74: A better name for the test could be `CreatingTrainerWithInvalidParameters_ShouldThrowException`.
You are testing with multiple invalid parameters, not just `userId`.
The name also better indicates that an exception is thrown.

In general, I prefer naming tests in the format of `DoingStuff_ShouldHaveThisResult`.

Note: this naming remark also applies to the other tests.

Smart.FA.Catalog.Tests.Common/TrainerFactory.cs
```diff
@@ -1,26 +1,62 @@
...
 using CSharpFunctionalExtensions;
 
 namespace Smart.FA.Catalog.Tests.Common;
 
-public class TrainerFactory
+public static class TrainerFactory
 {
-    public Trainer CreateClean()
+    private static Fixture fixture = new();
+    public static Trainer CreateClean()
```
LievenPraet, line 13: This function can be named just `Create`.
I don't think the word clean adds much value here.

Smart.FA.Catalog.Tests.Common/TrainerFactory.cs
```diff
@@ -1,26 +1,62 @@
...
+        return new Trainer
+        (
+            name.Value
+            , TrainerIdentity.Create(fixture.Create<string>()
+                , ApplicationType.Account).Value
+            , fixture.Create<string>()
+            , fixture.Create<string>(), defaultLanguage.Value);
     }
 
+    public static Trainer CreateFromUser(UserDto user)
```
LievenPraet, line 46: This function can also be named `Create` in which case, an overload of the function will be created with different parameters.

Smart.FA.Catalog.Tests.Common/TrainingFactory.cs
```diff
@@ -5,11 +5,12 @@
 
 namespace Smart.FA.Catalog.Tests.Common;
 
-public class TrainingFactory
+public static class TrainingFactory
 {
-    public Training Create(Trainer trainer)
+    private static Fixture fixture = new();
+
+    public  static Training Create(Trainer trainer)
```
LievenPraet, line 12: If the function accepts a trainer that can be null, the type should be nullable -> `Trainer?`
If null is not acceptable, an `ArgumentNullException` should be thrown.

Smart.FA.Catalog.Tests.Common/TrainingFactory.cs
```diff
@@ -21,15 +22,13 @@ public Training Create(Trainer trainer)
...
-    public Training CreateClean()
+    public static Training CreateClean()
     {
-        TrainerFactory trainerFactory = new();
-        return Create(trainerFactory.CreateClean());
+        return Create(TrainerFactory.CreateClean());
     }
 
-    public Training CreateWithManualValidation(Trainer trainer)
+    public static Training CreateWithManualValidation(Trainer trainer)
```
LievenPraet, line 30: Why does this function exist?
There is no difference with the function `Create`.

Smart.FA.Catalog.Tests.Common/TrainingFactory.cs
```diff
@@ -41,9 +40,8 @@ public Training CreateWithManualValidation(Trainer trainer)
         );
     }
 
-    public Training CreateWithAutoValidation(Trainer trainer)
+    public static Training CreateWithAutoValidation(Trainer trainer)
```
LievenPraet, line 43: Same remark as above.

Smart.FA.Catalog.Tests.Common/UserFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+public static class UserFactory
+{
+    private static Fixture fixture = new();
+
+    public static UserDto CreateClean()
+    {
+        return new UserDto
+        (
+            fixture.Create<int>()
+                .ToString(CultureInfo.CurrentCulture)
```
LievenPraet, line 17: Why include the culture here?
It makes no difference for an `int` like it would for a `decimal` or `float`.

Smart.FA.Catalog.UnitTests/Trainer_Tests.cs
```diff
@@ -11,14 +11,12 @@ namespace Smart.FA.Catalog.UnitTests;
 
 public class TrainerTests
 {
-    private readonly TrainerFactory _trainerFactory = new();
-    private readonly TrainingFactory _trainingFactory = new();
     private readonly Fixture _fixture = new();
 
     [Fact]
     public void HasInitiallyNoTraining()
```
LievenPraet, line 17: Rename to `CreatingTraining_ShouldNotHaveAssignments`

PR #31, opened 2022-03-21
Catalog Context refactoring

Victor-van-Duijnen:
I introduce some small refactoring. 
Since they are very small, instead of spamming you with 3 different PR's, I have madeonly  one PR with some small commits.
The PR includes:

1. Replace hard dependencies in the constructor of the catalog context with an injected option pattern
2. Renaming of the method injecting a catalog context in the integration tests

### DIFF
--- a/Smart.FA.Catalog.IntegrationTests/Base/ConnectionSetup.cs
+++ b/Smart.FA.Catalog.IntegrationTests/Base/ConnectionSetup.cs
@@ -20,7 +20,7 @@ public class ConnectionSetup
         new SqlConnectionStringBuilder
         {
             DataSource = @"(LocalDB)\MSSQLLocalDB",
-            InitialCatalog = "Training",
+            InitialCatalog = "Catalog",
             IntegratedSecurity = true
         };
 
--- a/Smart.FA.Catalog.IntegrationTests/TrainingContext/User_Tests.cs
+++ b/Smart.FA.Catalog.IntegrationTests/TrainingContext/User_Tests.cs
@@ -1,12 +1,12 @@
 ﻿using System.Threading;
 using System.Threading.Tasks;
-using Application.UseCases.Queries;
 using AutoFixture;
-using Core.Domain.Enumerations;
 using FluentAssertions;
-using Infrastructure.Services;
 using Microsoft.Extensions.Logging;
 using NSubstitute;
+using Smart.FA.Catalog.Application.UseCases.Queries;
+using Smart.FA.Catalog.Core.Domain.User.Enumerations;
+using Smart.FA.Catalog.Infrastructure.Services;
 using Xunit;
 using Smart.FA.Catalog.IntegrationTests.Base;
 
--- a/Smart.FA.Catalog.IntegrationTests/TrainingContext/User_Tests.cs
+++ b/Smart.FA.Catalog.IntegrationTests/TrainingContext/User_Tests.cs
@@ -20,7 +20,7 @@ public class UserTests: IntegrationTestBase
     public async Task CanGet()
     {
         var logger = Substitute.For<ILogger<GetTrainerFromUserAppQueryHandler>>();
-        await using var context = GivenTrainingContext();
+        await using var context = GivenCatalogContext();
         var userStrategyResolver = Substitute.For<UserStrategyResolver>("");
         var handler = Substitute.For<GetTrainerFromUserAppQueryHandler>(logger, context, userStrategyResolver );
         var request = new GetTrainerFromUserAppRequest {ApplicationType = ApplicationType.Account, UserId = _fixture.Create<string>()};
--- a/Smart.FA.Catalog.IntegrationTests/TrainingContext/User_Tests.cs
+++ b/Smart.FA.Catalog.IntegrationTests/TrainingContext/User_Tests.cs
@@ -1,0 +1,33 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+using Application.UseCases.Queries;
+using AutoFixture;
+using Core.Domain.Enumerations;
+using FluentAssertions;
+using Infrastructure.Services;
+using Microsoft.Extensions.Logging;
+using NSubstitute;
+using Xunit;
+using Smart.FA.Catalog.IntegrationTests.Base;
+
+namespace Smart.FA.Catalog.IntegrationTests.TrainingContext;
+[Collection("Integration test collection")]
+public class UserTests: IntegrationTestBase
+{
+    private readonly Fixture _fixture = new();
+
+    [Fact]
+    public async Task CanGet()
+    {
+        var logger = Substitute.For<ILogger<GetTrainerFromUserAppQueryHandler>>();
+        await using var context = GivenTrainingContext();
+        var userStrategyResolver = Substitute.For<UserStrategyResolver>("");
+        var handler = Substitute.For<GetTrainerFromUserAppQueryHandler>(logger, context, userStrategyResolver );
+        var request = new GetTrainerFromUserAppRequest {ApplicationType = ApplicationType.Account, UserId = _fixture.Create<string>()};
+
+        var user = await handler.Handle(request, CancellationToken.None);
+
+        user.Should().NotBeNull();
+        user.User.UserId.Should().Equals(request.UserId);
+    }
+}
--- a/Smart.FA.Catalog.IntegrationTests/TrainingContext/User_Tests.cs
+++ b/Smart.FA.Catalog.IntegrationTests/TrainingContext/User_Tests.cs
@@ -17,7 +17,7 @@ public class UserTests: IntegrationTestBase
     private readonly Fixture _fixture = new();
 
     [Fact]
-    public async Task CanGet()
+    public async Task GetTrainerFromUserAppRequest()
     {
         var logger = Substitute.For<ILogger<GetTrainerFromUserAppQueryHandler>>();
         await using var context = GivenCatalogContext();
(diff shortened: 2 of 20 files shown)


Smart.FA.Catalog.IntegrationTests/TrainingContext/User_Tests.cs
```diff
@@ -0,0 +1,33 @@
...
+    [Fact]
+    public async Task CanGet()
+    {
+        var logger = Substitute.For<ILogger<GetTrainerFromUserAppQueryHandler>>();
+        await using var context = GivenCatalogContext();
+        var userStrategyResolver = Substitute.For<UserStrategyResolver>("");
+        var handler = Substitute.For<GetTrainerFromUserAppQueryHandler>(logger, context, userStrategyResolver );
+        var request = new GetTrainerFromUserAppRequest {ApplicationType = ApplicationType.Account, UserId = _fixture.Create<string>()};
+
+        var user = await handler.Handle(request, CancellationToken.None);
```
LievenPraet, line 28: Var should be named `trainer` 
```suggestion
        var trainer = await handler.Handle(request, CancellationToken.None);
```

Smart.FA.Catalog.IntegrationTests/TrainingContext/User_Tests.cs
```diff
@@ -0,0 +1,33 @@
...
+using Smart.FA.Catalog.IntegrationTests.Base;
+
+namespace Smart.FA.Catalog.IntegrationTests.TrainingContext;
+[Collection("Integration test collection")]
+public class UserTests: IntegrationTestBase
+{
+    private readonly Fixture _fixture = new();
+
+    [Fact]
+    public async Task CanGet()
```
LievenPraet, line 20: We are calling `GetTrainerFromUserAppRequest` but the function name is `CanGet`.
The name of this function is not clear and doesn't give enough / correct details of what we are testing.

src/Smart.FA.Catalog.Application/UseCases/DomainEventHandlers/ValidateTrainingDomainEventHandler.cs
```diff
@@ -0,0 +1,23 @@
...
+
+    public ValidateTrainingDomainEventHandler(MessageBus messageBus)
+    {
+        _messageBus = messageBus;
+    }
+
+    public Task Handle(ValidateTrainingEvent notification, CancellationToken cancellationToken)
+    {
+        _messageBus.ValidateTrainingMessage(notification.TrainingId,
+            notification.TrainingName, notification.TrainersId);
```
LievenPraet, line 19: Will an exception be thrown if the validation fails?

src/Smart.FA.Catalog.Infrastructure/DesignTimeContextFactory.cs
```diff
@@ -1,29 +1,33 @@
...
+
 public class DesignTimeContextFactory : IDesignTimeDbContextFactory<CatalogContext>
 {
     public CatalogContext CreateDbContext(string[]? args)
     {
         var basePath = Directory.GetCurrentDirectory();
         var environmentName = Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT");
-        Console.WriteLine($"\nDesignTimeContextFactory.Create(string[]):\n\tBase Path: {basePath}\n\tEnvironmentVariable: {environmentName}");
+        Console.WriteLine(
+            $"\nDesignTimeContextFactory.Create(string[]):\n\tBase Path: {basePath}\n\tEnvironmentVariable: {environmentName}");
```
LievenPraet, line 18: Could we use a logger here instead of console writeline?

LievenPraet (changes requested):

Smart.FA.Catalog.IntegrationTests/TrainingContext/User_Tests.cs
Victor-van-Duijnen: Yes good point

src/Smart.FA.Catalog.Infrastructure/DesignTimeContextFactory.cs
Victor-van-Duijnen: The DesignTimeContextFactory is only used when you want to recreate the state of the DB from ef tools commands inside the project (in isolation of the solution) so actually no debugging need to/should be done in that part (the console message is purely informative about the progress of the designtimecontextfactory). It is NEVER used at runtime : https://docs.microsoft.com/en-us/ef/core/cli/dbcontext-creation?tabs=dotnet-core-cli#from-a-design-time-factory

Smart.FA.Catalog.IntegrationTests/TrainingContext/User_Tests.cs
Victor-van-Duijnen: Yes I can see how that can be confusing, I will change it. Right now my naming convention would have been 'CanGetFromTrainer' (the scope of the tests is defined by the name of the class, here User_Tests, so all operations are definitively made on the User). The 'Can' part of the message indicates that all input of the test should be valid (to differentiate 'CantGet' tests when we have invalid input). But I will standardize my naming of tests to yours (starting by that one)!

src/Smart.FA.Catalog.Application/UseCases/DomainEventHandlers/ValidateTrainingDomainEventHandler.cs
Victor-van-Duijnen: That method is not validating anything, it's just propagating the domain event (the validating event itself is done before and the exception handling, too). If the validating fails, it will indeed throw a "Validating error" and will never arrives in the method described here.  That method here is just supposed to log some message right now and that's why there is no exception handling.

maxime-poulain: Regardless you go for event driven architecture or not, publishing events should be restricted to strong state changes. Namely creation, edition, stuff like that.
You might prefer to have something that surrounds every validation and do something upon validating rather than using an event.

Victor-van-Duijnen: No, your a confusing what I do in that validation. That validation is for switching a training from status "In Review" or "In Draft" to a status Published, not to validate if the model is valid. So it's a very important state change that definitively needs to be published as an event.

src/Smart.FA.Catalog.Infrastructure/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -72,4 +69,31 @@ private static IServiceCollection AddQueries(this IServiceCollection services, s
...
+                builder.AddFilter((category, level) =>
+                        category == DbLoggerCategory.Database.Command.Name && level == LogLevel.Information)
+                    .AddConsole();
+            });
+
+            options
+                .UseSqlServer(connectionString)
+                .UseLazyLoadingProxies();
+            if (dalOptions.Value.UseConsoleLogger)
+            {
```
maxime-poulain, line 92: I guess you can neste the loggerFactory create inside this ef.
Wouldn't it be better to set the log level through the app settings rather than hardcoded that here ?
`    "LogLevel": {
      "Default": "Information",
      "Microsoft": "Warning",
      "Microsoft.Hosting.Lifetime": "Information"
    }`

src/Smart.FA.Catalog.Web/appsettings.Development.json
```diff
@@ -20,5 +20,9 @@
     "Training": {
       "NumberOfTrainingsDisplayed": 3
     }
+  },
+  "DALOptions":
+  {
```
maxime-poulain, line 25: Hm I think you can remove the Options suffix. "Logging" section is also behind the scene an option for example;
Would it  make sense to name it also name it someting like EfCore as this settings is strongly related to that technology ?

src/Smart.FA.Catalog.Infrastructure/Extensions/ServiceCollectionExtensions.cs
Victor-van-Duijnen: Standard logs in the application should be separated from raw queries (useconsolelogger) only used for very specific debugging purpose. They should only be used in particular scenario where it would give important tracing information, but never for a long period time since they are VERY VERY verbose (even more than the trace log level) obviously. It already depends on the appsettings and it's therefore not hardcoded (but I guess you meant the fact that it is a cutsom field rather than the customary log settings). I suggest we keep it separated from the standard logs.

src/Smart.FA.Catalog.Web/appsettings.Development.json
Victor-van-Duijnen: Ok I can do that, II have tended to use the "Options" postfix for all information but maybe it's a bit redundant indeed. I will change it.

Smart.FA.Catalog.IntegrationTests/Base/ConnectionSetup.cs
```diff
@@ -20,7 +20,7 @@ public class ConnectionSetup
         new SqlConnectionStringBuilder
         {
             DataSource = @"(LocalDB)\MSSQLLocalDB",
-            InitialCatalog = "Training",
+            InitialCatalog = "Catalog",
```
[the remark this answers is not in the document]
Victor-van-Duijnen: It's true we have decided to go for a more simple, straightforward name for our DBs but we are open to suggestions!

src/Smart.FA.Catalog.Infrastructure/Extensions/ServiceCollectionExtensions.cs
maxime-poulain: Well, what I meant, was "is it possible to filter the console logs with a appsettings:" ?
 let's say you add this in your appsetting:
`{
"LogLevel": { "Microsoft.EntityFrameworkCore.Database.Command": "Trace"
}`

than commands won't be written however the reste will remains at the information level.
The idea was to know if we can setting up, in the same way I described above.
The idea is that locally let's say I want to have some output in the console but not the command, is it possible ? 
Writing logs to a file is way to retarded,I was not suggesting that :p. There would be too many I/O.

>I guess you can neste the loggerFactory create inside this ~~ef~~ **if**, sorry.

PR #33, opened 2022-03-22
Change all namespaces to include full name 

Victor-van-Duijnen:
This PR is impacting  large variety of files (close to all of them) because it's about changing the namespace convention.
Full name of the project is now the root namespace of every project. Smart.FA.Catalog.{Project}.

If that PR could be validated quickly so that I can avoid conflicts for future PR, please!


maxime-poulain (approved):

Smart.FA.Catalog.sln
```diff
@@ -32,11 +32,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Core", "src\Smart.FA.Catalo
...
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Smart.FA.Catalog.IntegrationTests", "Smart.FA.Catalog.IntegrationTests\Smart.FA.Catalog.IntegrationTests.csproj", "{BBE101FD-3BFE-40C6-8ABA-5D5C268541EF}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "IntegrationTests", "Smart.FA.Catalog.IntegrationTests\IntegrationTests.csproj", "{BBE101FD-3BFE-40C6-8ABA-5D5C268541EF}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Smart.FA.Catalog.UnitTests", "Smart.FA.Catalog.UnitTests\Smart.FA.Catalog.UnitTests.csproj", "{B8F4E02A-C635-4729-8E94-117F4886BE65}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "UnitTests", "Smart.FA.Catalog.UnitTests\UnitTests.csproj", "{B8F4E02A-C635-4729-8E94-117F4886BE65}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Smart.FA.Catalog.Tests.Common", "Smart.FA.Catalog.Tests.Common\Smart.FA.Catalog.Tests.Common.csproj", "{BEB6142A-19A0-410D-BA50-1E0DD035A643}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Tests.Common", "Smart.FA.Catalog.Tests.Common\Tests.Common.csproj", "{BEB6142A-19A0-410D-BA50-1E0DD035A643}"
```
maxime-poulain, line 39: Every project's name should be the same than the csproj file.
Displaying just  Application / Core / etc is not enough.
I've been there at some point; I was not understanding why people would just put the full root namespace as a name.
Then I realized it was not a good practice actually.

Smart.FA.Catalog.UnitTests/UnitTests.csproj
```diff
@@ -5,6 +5,10 @@
     <Nullable>enable</Nullable>
 
     <IsPackable>false</IsPackable>
+
+    <RootNamespace>Smart.FA.Catalog.UnitTests</RootNamespace>
+
+    <AssemblyName>Smart.FA.Catalog.UnitTests</AssemblyName>
   </PropertyGroup>
```
maxime-poulain, line 12: You could even remove those settings, it takes by default the csproj name for those value.

LievenPraet (approved):

maxime-poulain:
LIke I mentioned here 
https://github.com/smartcoop/smart-cfa/pull/33/files/763948fc7978586a27f1b6b3f1392acad15ae28e#r832089931

Project's name should be equals to the csproj / root namespace.

This is the most common way of doing one can find.

Also as @LievenPraet  mentioned if we get multiple project it is unclear what is what.

PR #34, opened 2022-03-22
Event dispatcher refactoring

Victor-van-Duijnen:
dispatching domain event incumbents now to the mediator (it publishes events picked up that are by specific domain event handlers)


maxime-poulain:
@LievenPraet  we won't go with this impl.

maxime-poulain:
Replaced by
https://github.com/smartcoop/smart-cfa/pull/35

PR #36, opened 2022-03-22
Introduce localization for CFA

maxime-poulain:
This introduces support for localization.
Is supported FR,NL,EN.
We will be setting all translation inside the same resource file.
Therefore it is encouraged to use the underscore to scope some keys
example
>TrainingCreationPage_Title that would be "This is the super title of a training".
>TriainingTitle should be used for Training title

We will see how the naming evolve and make out of it something consistent.


maxime-poulain:
Ideally this should be merged asap.

Victor-van-Duijnen (approved):

PR #57, opened 2022-03-30
MediatR: when serializing requests prevent circular errors

maxime-poulain:
The LoggingPipelineBehavior class logs in development environments requests that come in.
If the request has circular references, example, an entity, such Trainer, it will result with into an exception because it produces an infinite loop.

Internal Jira ticket: https://smartjira.atlassian.net/browse/CFA-290

Some doc: https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-preserve-references?pivots=dotnet-6-0

### DIFF
--- a/src/UserAdmin/src/Smart.FA.Catalog.Application/Interceptors/LoggingPipelineBehavior.cs
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Application/Interceptors/LoggingPipelineBehavior.cs
@@ -1,3 +1,5 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
 using MediatR;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
@@ -41,8 +43,9 @@ public class LoggingPipelineBehavior<TRequest, TResponse> : IPipelineBehavior<TR
     {
         if (_mediatROptions.LogRequests)
         {
-            _logger.LogInformation("Incoming request: {requestName} - {request}", request.GetType().Name, request.ToJson());
+            LogMediatRRequest(request);
         }
+
         TResponse response = new();
         try
         {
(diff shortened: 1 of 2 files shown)


maxime-poulain:
@Victor-van-Duijnen 
The first time I encountered that what because of [this](https://github.com/smartcoop/smart-cfa/blob/7247a632c399bdec6696bce660c8184b917b6b1f/src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Commands/UploadTrainerProfileImageCommand.cs#L52).
This could have happened with something than an entity, but I think sending an entity from the web  layer to another is a bad practice. In my implementation I use the id rather the object itself.
What do you think ?

Victor-van-Duijnen:
> sending an entity from the web layer to another is a bad practice
@maxime-poulain Je suis assez d'accord avec toi, en fait il faut remonter à la source du problème, je pense pas qu'il faudrait qu'on aie un Trainer domain model au niveau de l'identity de l'user (plutôt un bête DTO qui contient le trainer id) pour éviter ce genre de probs dans le futur, ca va affecter pas mal de méthodes mais je vais regarder  pour ajouter ça dans une nouvelle PR si t'es OK avec ça.

maxime-poulain:
> > "sending an entity from the web layer to another is a bad practice"
> > @maxime-poulain
> > Do you rather mean than sending directly  an entity from the web layer to a handler is a bad practice? It's not communicating with another class on the same layer but if you meant the former, I agree with you.
> > We might directly fix the root of the problem which is that we are using a trainer domain model directly inside the identity of the user , don't you think it would even be better to only have a dumb Trainer dto with only the trainerId (and maybe a Name value object, too) instead? Although it will impact a few handler calls, if you agree, I want to introduce that in a new PR.

Well the web layer is encapsulating the entity inside an object that is sent to another layer it doesn't matter in the end how it is being sent (through a server or mediatr or w/e).

Currently the service I made exposes already the Id  https://github.com/smartcoop/smart-cfa/blob/b2e9d0cfc183dacf608f760ed34bd7ebfc346328/src/UserAdmin/src/Smart.FA.Catalog.Core/Services/IUserIdentity.cs#L14. I just kept the entity object not to refactor the whole stuff like I said back then. Indeed, a better approach would be to remove the entity and expose the infos we need.

So  it can be  more informations such the name, or email, or I don"t know. 
I assume those would be get only properties on the claims, could be a more sophisticated object  ?
In my cases, so far, having the first name / lastname and Id of the TrainerId and UserId is enough.

Victor-van-Duijnen (approved):

PR #63, opened 2022-04-06
Add practical modalities field to training

Victor-van-Duijnen:
1. Create a new field in the training model for the practical modalities with:

- Testing
- Validation
- Migration

2. Revise Precision in validators and domain layer of the training fields

- min and max length of Title, Methodology, Goal, PracticalModalities

### DIFF
--- a/src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Training/TrainingDetail.cs
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Training/TrainingDetail.cs
@@ -26,7 +26,7 @@ public class TrainingDetail
         set
         {
             Guard.AgainstNull(value, nameof(Title));
-            ValidateMaxLength(value, nameof(Methodology), 500);
+            Guard.AgainstMaxStringLength(value, nameof(Title), 500);
             _title = value;
         }
     }
@@ -36,7 +36,11 @@ public class TrainingDetail
         get => _goal;
         set
         {
-            ValidateMaxLength(value, nameof(Methodology), 1500);
+            if (value is not null)
+            {
+                Guard.AgainstMaxStringLength(value, nameof(Goal), 1000);
+            }
+
             _goal = value;
         }
     }
@@ -46,7 +50,11 @@ public class TrainingDetail
         get => _methodology;
         set
         {
-            ValidateMaxLength(value, nameof(Methodology), 1500);
+            if (value is not null)
+            {
+                Guard.AgainstMaxStringLength(value, nameof(Methodology), 1000);
+            }
+
             _methodology = value;
         }
     }
@@ -56,7 +64,11 @@ public class TrainingDetail
         get => _practicalModalities;
         set
         {
-            ValidateMaxLength(value, nameof(PracticalModalities), 1500);
+            if (value is not null)
+            {
+                Guard.AgainstMaxStringLength(value, nameof(PracticalModalities), 1000);
+            }
+
             _practicalModalities = value;
         }
     }
@@ -93,19 +105,6 @@ public class TrainingDetail
         Methodology = methodology;
         PracticalModalities = practicalModalities;
     }
-
-    #endregion
-
-    #region Validation
-
-    private void ValidateMaxLength(string? field, string? name, int maxValue) =>
-        LengthMaxValidation.Compile()(field, name, maxValue);
-
-    private static readonly Expression<Action<string?, string?, int>> LengthMaxValidation =
-        (fieldValue, fieldName, maxValue) =>
-            Guard.Requires(() =>
-                    string.IsNullOrEmpty(fieldValue) || fieldValue.Length < maxValue,
-                $"{nameof(fieldName)} has a maximum value of {maxValue} characters");
 
     #endregion
 }
--- a/src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Training/TrainingDetail.cs
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Training/TrainingDetail.cs
@@ -11,6 +11,7 @@ public class TrainingDetail
     private string _title = null!;
     private string? _goal;
     private string? _methodology;
+    private string? _practicalModalities;
 
     #endregion
 
@@ -50,13 +51,23 @@ public class TrainingDetail
         }
     }
 
+    public string? PracticalModalities
+    {
+        get => _practicalModalities;
+        set
+        {
+            ValidateMaxLength(value, nameof(PracticalModalities), 1500);
+            _practicalModalities = value;
+        }
+    }
+
     public Language Language { get; } = null!;
 
     #endregion
 
     #region Constructors
 
-    internal TrainingDetail(Training training, string title, string? goal, string? methodology, Language language)
+    internal TrainingDetail(Training training, string title, string? goal, string? methodology, string? practicalModalities, Language language)
     {
         Training = training;
         TrainingId = training.Id;
@@ -64,6 +75,7 @@ public class TrainingDetail
         Goal = goal;
         Methodology = methodology;
         Language = language;
+        PracticalModalities = practicalModalities;
     }
 
     protected TrainingDetail()
@@ -74,11 +86,12 @@ public class TrainingDetail
 
     #region Methods
 
-    public void UpdateDescription(string title, string goal, string methodology)
+    public void UpdateDescription(string title, string? goal, string? methodology, string? practicalModalities)
     {
         Title = title;
         Goal = goal;
         Methodology = methodology;
+        PracticalModalities = practicalModalities;
     }
 
     #endregion
@@ -91,8 +104,8 @@ public class TrainingDetail
     private static readonly Expression<Action<string?, string?, int>> LengthMaxValidation =
         (fieldValue, fieldName, maxValue) =>
             Guard.Requires(() =>
-                    string.IsNullOrEmpty(fieldValue) || fieldValue.Length < 1500,
-                $"{nameof(fieldName)} has a maximum value of 1500 characters");
+                    string.IsNullOrEmpty(fieldValue) || fieldValue.Length < maxValue,
+                $"{nameof(fieldName)} has a maximum value of {maxValue} characters");
 
     #endregion
 }
--- a/src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Training/TrainingDetail.cs
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Training/TrainingDetail.cs
@@ -26,51 +26,26 @@ public class TrainingDetail
         set
         {
             Guard.AgainstNull(value, nameof(Title));
-            Guard.AgainstMaxStringLength(value, nameof(Title), 500);
-            _title = value;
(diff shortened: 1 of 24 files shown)


src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Validators/TrainingDetailValidation.cs
```diff
@@ -15,5 +15,7 @@ public TrainingDetailValidation()
             .NotEmptyWithGenericMessage().WithMessage(Errors.General.MissingField("goal").Message);
         RuleFor(request => request.Methodology)
             .NotEmptyWithGenericMessage().WithMessage(Errors.General.MissingField("methodology").Message);
+        RuleFor(request => request.PracticalModalities)
+            .NotEmptyWithGenericMessage().WithMessage(Errors.General.MissingField("practicalModalities").Message);
```
maxime-poulain, line 19: Why not use the resources here ?
I assume those validation messages are used somehow ?

src/UserAdmin/src/Smart.FA.Catalog.Localization/CatalogResources.resx
```diff
@@ -282,6 +285,9 @@
   <data name="Min30CharMax500Char" xml:space="preserve">
     <value>min 30 caractères / max 500 caractères</value>
   </data>
+  <data name="Min30CharMax1000Char" xml:space="preserve">
+    <value>min 30 caractères / max 1000 caractères</value>
+  </data>
```
maxime-poulain, line 290: You use in the above code 1500 for the max length.

maxime-poulain (changes requested):

maxime-poulain:
Hello,
While reviewing this I noticed that the validators of Models class  name is different than the file name but I cant write a comment sadly in the code :(
https://github.com/smartcoop/smart-cfa/blob/7247a632c399bdec6696bce660c8184b917b6b1f/src/UserAdmin/src/Smart.FA.Catalog.Web/Validators/CreateTrainingRequestValidator.cs

src/UserAdmin/src/Smart.FA.Catalog.Web/Validators/CreateTrainingRequestValidator.cs
```diff
@@ -43,6 +43,14 @@ private void ApplyRemainingRules()
             .MaximumLength(500)
             .WithMessage(" ");
 
+        RuleFor(viewModel => viewModel.PracticalModalities)
+            .NotEmpty()
+            .WithMessage(CatalogResources.FieldRequired)
+            .MinimumLength(30)
+            .WithMessage(CatalogResources.Min30CharMax1000Char)
+            .MaximumLength(1000)
```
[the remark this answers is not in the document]
Victor-van-Duijnen: Well, yes, I did'nt want to introduce that in that PR because it needs change on whole the fields of the object (it's in my next PR). When you defined the min Lenght and max length in validator without changing the business logic, it was already wrong but I have only noticed it when adding the field.

src/UserAdmin/src/Smart.FA.Catalog.Localization/CatalogResources.resx
Victor-van-Duijnen: No, not in the validator of the pages. I intend to create a PR with the right field size for multiple fields in the database that were not in phase with what was added in the validators but it's outside the scope of that PR.

src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Validators/TrainingDetailValidation.cs
Victor-van-Duijnen: This the validation used in the Domain layer. It's not meant to be displayed but logged so It should not be localized.

src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Training/TrainingDetail.cs
```diff
@@ -91,8 +104,8 @@ public void UpdateDescription(string title, string goal, string methodology)
     private static readonly Expression<Action<string?, string?, int>> LengthMaxValidation =
         (fieldValue, fieldName, maxValue) =>
             Guard.Requires(() =>
-                    string.IsNullOrEmpty(fieldValue) || fieldValue.Length < 1500,
-                $"{nameof(fieldName)} has a maximum value of 1500 characters");
+                    string.IsNullOrEmpty(fieldValue) || fieldValue.Length < maxValue,
```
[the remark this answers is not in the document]
Victor-van-Duijnen: The initial idea was to use expressions to create lots of dynamic filters with EF (not possible with Action) but I don't think I will go that road anyway, so I will refactor that.

Victor-van-Duijnen:
Following @maxime-poulain 's comments I have extended the scope of the PR to include revision of the precision (min/max length) of the training fields (Title, Goal, Methodology and PracticalModalities).

src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Validators/TrainingDetailValidator.cs
```diff
@@ -3,9 +3,9 @@
 
 namespace Smart.FA.Catalog.Core.Domain.Validators;
 
-public class TrainingDetailValidation: AbstractValidator<TrainingDetail>
+public class TrainingDetailValidator: AbstractValidator<TrainingDetail>
```
maxime-poulain, line 6:
```suggestion
public class TrainingDetailValidator : AbstractValidator<TrainingDetail>
```

src/UserAdmin/src/Smart.FA.Catalog.Core/SeedWork/Guard.cs
```diff
@@ -34,4 +34,19 @@ public static string AgainstInvalidEmail(string? email, string parameterName, st
...
     }
+
+    public static void AgainstMaxStringLength(string value, string valueName, int maxValue, string? message = null)
+    {
+        if (value.Length > maxValue)
+        {
+            throw new ArgumentException(message ?? $"{valueName} needs a maximum length of {maxValue} characters");
+        }
+    }
+    public static void AgainstMinStringLength(string value, string valueName, int minValue, string? message = null)
```
maxime-poulain, line 45: missing breakline

src/UserAdmin/src/Smart.FA.Catalog.Core/SeedWork/Guard.cs
```diff
@@ -34,4 +34,19 @@ public static string AgainstInvalidEmail(string? email, string parameterName, st
 
         return email!;
     }
+
+    public static void AgainstMaxStringLength(string value, string valueName, int maxValue, string? message = null)
```
maxime-poulain, line 38: Donc include "String" in your method name it can be inferred easily thanks to the value parameter.

Later on there will be most likely another method on collection's maximum, we dont need to have something called
>AgainstMaxCollectionLength

I would rename valueName to "parameterName" or "argumentName" as it refers to the calling parameter.
maxValue to maximumLength to make it less vague.

Regarding the method name I think AgainstMaximumLength makes more sense than AgainstMaxLength

src/UserAdmin/src/Smart.FA.Catalog.Web/Validators/CreateTrainingRequestValidator.cs
```diff
@@ -31,17 +47,19 @@ private void ApplyRemainingRules()
             .NotEmpty()
             .WithMessage(CatalogResources.FieldRequired)
             .MinimumLength(30)
-            .WithMessage(" ")
-            .MaximumLength(500)
-            .WithMessage(" ");
+            .WithMessage(CatalogResources.Min30Char);
```
maxime-poulain, line 51: I did not put any error message because we have already on the helper text telling us the min / max length information.
I think that having twice the same message is not really user friendly.

src/UserAdmin/src/Smart.FA.Catalog.Web/Validators/UpdateTrainingRequestValidator.cs
```diff
@@ -15,6 +15,21 @@ public UpdateTrainingViewModelValidator(IStringLocalizer<CatalogResources> local
             .NotEmpty()
             .WithMessage(CatalogResources.TrainingTitleIsRequired);
 
+        When(viewModel => viewModel.Methodology is not null, () =>
+            RuleFor(viewModel => viewModel.Methodology)
```
maxime-poulain, line 19: FluentValidation's length valiator doesn't care if the value if null.
See
https://github.com/FluentValidation/FluentValidation/blob/2671e7fb09af9e1ef2a938a8b4364724ef9e47bd/src/FluentValidation/Validators/LengthValidator.cs#L47

Victor-van-Duijnen: Nice spotting! I will remove the redundant where clause.

src/UserAdmin/src/Smart.FA.Catalog.Web/Validators/CreateTrainingRequestValidator.cs
Victor-van-Duijnen: Hmm not totally agree, you might not automatically know if you have reached 30 characters, so you have to know with a error text that the field is not valid and need changes.

maxime-poulain: Thinking of it again yeah should be better to have an error message even if e have an helper text.

However, why did you remove the max length on the validator when publishing ?
Isn't there a limit expected ? I know we do at a database constraint at least.

Victor-van-Duijnen: No no maxLength is applied but since  the part commented, is limited to published training and that the limit of character should be applied to both draft and published training, I have moved that logic up (if you look at the start if the validator you can see that there is length max validation happening there)

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Persistence/Configuration/TrainingDetailConfigurations.cs
```diff
@@ -15,10 +15,13 @@ public void Configure(EntityTypeBuilder<TrainingDetail> builder)
...
         builder.Property(detail => detail.Goal)
-            .HasMaxLength(1500)
+            .HasMaxLength(1000)
             .IsRequired(false);
         builder.Property(detail => detail.Methodology)
-            .HasMaxLength(1500)
+            .HasMaxLength(1000)
+            .IsRequired(false);
+        builder.Property(detail => detail.PracticalModalities)
+            .HasMaxLength(1000)
```
maxime-poulain, line 24: I think the guard clause + db constraint is 1500

Victor-van-Duijnen: Hmm nop everything is 1000 (Guard + DB + Validator) but I agree I had some doubts about it when I first read your comment :b

maxime-poulain (approved):

PR #64, opened 2022-04-06
Feature/cfa 250 homepage template

joe-ubik:
I had a conflict on this branch. Have resolve it. Not sure the process was good


nyvio:
Coucou @joe-ubik fais-moi signe lorsque tu auras pushé les dernières modifs et que je peux faire la review.

nyvio (approved):

PR #66, opened 2022-04-07
Fix JS error on loading of most pages

maxime-poulain:
There was  a onload event logic applied globally to every pages but that logic is related only to creation and edition pages.
Therefore opening any pages but those two would result in this:
'Cannot read properties of null (reading 'addEventListener')'.
This PR provides a quick and dirty fix by returning in nothing is found.

Internal ticket:
https://smartjira.atlassian.net/browse/CFA-322

### DIFF
--- a/src/UserAdmin/src/Smart.FA.Catalog.Web/wwwroot/js/lib.js
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Web/wwwroot/js/lib.js
@@ -1,7 +1,13 @@
 $(document).ready(
+    // This is used only by the creation and edition page therefore it could be move somewhere else.
+    // If one has the heart to do so he/she is welcome :).
+    // Btw try avoiding jQuery please :(.
     function(event)
     {
         let validateButton = document.getElementById("Save");
+        if (!validateButton) {
+           // return;
+        }
         validateButton.addEventListener("click", function(e)
         {
             let model = document.getElementById("isDraft");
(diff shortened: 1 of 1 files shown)


src/UserAdmin/src/Smart.FA.Catalog.Web/wwwroot/js/lib.js
```diff
@@ -1,7 +1,13 @@
 $(document).ready(
+    // This is used only by the creation and edition page therefore it could be move somewhere else.
```
Victor-van-Duijnen, line 2:
```suggestion
    // This is used only by the creation and edition page therefore it could be move somewhere else.
```
```suggestion
    // This is used only by the creation and edition page therefore it could be moved somewhere else.
```

src/UserAdmin/src/Smart.FA.Catalog.Web/wwwroot/js/lib.js
```diff
@@ -1,7 +1,13 @@
...
+    // If one has the heart to do so he/she is welcome :).
+    // Btw try avoiding jQuery please :(.
```
Victor-van-Duijnen, line 4: Let's try to avoid recommending guidelines directly in code comments. It could be discussed over informally together or more officially in a rulebook that we would create :)

Victor-van-Duijnen, line 4: Let's try to avoid recommending guidelines directly in code comments. It could be discussed over informally together or more officially in a rulebook that we would create :)

Victor-van-Duijnen (changes requested):

Victor-van-Duijnen:
Nice initiative!

src/UserAdmin/src/Smart.FA.Catalog.Web/wwwroot/js/lib.js
maxime-poulain: Ty

Victor-van-Duijnen (approved):

PR #67, opened 2022-04-08
Feature/cfa 269 add all trainings page

CelineMarique:
Showcase > Add razor page for all trainings display

https://smartjira.atlassian.net/browse/CFA-269

Adapt DbContext to query training view

    Adapt v_TrainingList to get training status and topic id instead of name.
    Add keys to DbContext and connection settings in DbContext and Program.cs

Add TrainingList page

    Add specific folder for trainings and create Traininglist page.

Add basic template for trainings list
Get sorted training list

    Add a training view model.
    Adapt page controller and html to get only validated trainings and display training topic id (the idea is to have a translation for the topic name at some point).

This is part of the story to add a pages all trainings page: https://smartjira.atlassian.net/browse/CFA-256


src/Showcase/src/Smart.FA.Catalog.Showcase.Infrastructure/DesignTimeContextFactory.cs
```diff
@@ -0,0 +1,16 @@
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Design;
+using Smart.FA.Catalog.Showcase.Infrastructure.Data;
+
+namespace Smart.FA.Catalog.Infrastructure;
+public class DesignTimeContextFactory : IDesignTimeDbContextFactory<CatalogShowcaseContext>
+{
```
maxime-poulain, line 7: hm did the tool generate that for you ?

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList.cshtml
```diff
@@ -0,0 +1,57 @@
+﻿@page
+@model Smart.FA.Catalog.Showcase.Web.Pages.Training.TrainingListModel
```
maxime-poulain, line 2: I would move the cshtml file and its viewmodel into a seperated folder but if you do so, you need to specify the route like

@page "training/list" or something similar.

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList.cshtml.cs
```diff
@@ -0,0 +1,68 @@
...
+        _context = context;
+    }
+
+    public IList<TrainingList> TrainingList { get; set; }
+    public List<Training> Trainings { get; set; } = new List<Training>();
+
+    public async Task OnGetAsync()
+    {
+        TrainingList = await _context.TrainingList.ToListAsync();
+        TrainingList = TrainingList.OrderBy(o => o.TrainingId).ToList();
```
maxime-poulain, line 24: You could make TrainingList a local variable and remove therefore the public property.
Also i would filter on the IQueryable the status that way to do only one call to ToList

```
var trainingList = _context.TrainingList.Where(training => training.TrainingStatus == TrainingStatus.Validated.Id).ToListAsync();
```

you could also remove the order by  linq extension method ?

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList.cshtml.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        foreach (var training in TrainingList)
+        {
```
maxime-poulain, line 27: This seems to me a bit complicated, that 
that should work also
```
        var groupedTrainingByIds = trainingList.ToLookup(t => t.TrainingId);
        var toto = groupedTrainingByIds.Select(trainings =>
        {
            // Since the source is a view, every items of the list has the same training properties.
            var training = trainings.First();
            return new Training()
            {
                TrainerFirstName = training.TrainerFirstName,
                TrainerLastName  = training.TrainerLastName,
                TrainingId       = training.TrainingId,
                TrainingStatus   = TrainingStatus.Validated,
                TrainingTitle    = training.TrainingTitle,
                Topics           = training.Select(x => TrainingTopic.FromValue<TrainingTopic>(x.TrainingTopic)).ToList()
            };
        });
 ```

File: src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Program.cs [hunk 5]
```diff
@@ -1,8 +1,17 @@
...
 
 // Add services to the container.
 builder.Services.AddRazorPages();
 
+builder.Services.AddDbContext<CatalogShowcaseContext>((_, efOptions) =>
+{
+    efOptions.UseSqlServer(builder.Configuration.GetConnectionString("Catalog"));
+});
+builder.Services.AddHttpContextAccessor();
+
```
maxime-poulain, line 14: This can be remove, it allows you to inject throught IHttpontextAccessor interface the current http context but in our case it wont be used, or at least i think.

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Program.cs
```diff
@@ -1,8 +1,17 @@
+using Microsoft.EntityFrameworkCore;
+using Smart.FA.Catalog.Showcase.Infrastructure.Data;
+
 var builder = WebApplication.CreateBuilder(args);
 
 // Add services to the container.
 builder.Services.AddRazorPages();
 
+builder.Services.AddDbContext<CatalogShowcaseContext>((_, efOptions) =>
+{
```
maxime-poulain, line 10: This works also

```suggestion
builder.Services.AddDbContext<CatalogShowcaseContext>(efOptions)
```

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/Training.cs
```diff
@@ -0,0 +1,14 @@
+using Smart.FA.Catalog.Shared.Domain.Enumerations.Training;
+
+namespace Smart.FA.Catalog.Showcase.Web.Pages.Training;
+
+public class Training
```
maxime-poulain, line 5: Some people like to append "viewModel" at the end of classes that are exposing informations to your page/view.

I think it could make more sense to have it as the class is not in some "ViewModel" folder but honestly up to you.

src/Showcase/src/Smart.FA.Catalog.Showcase.Domain/efpt.config.json
```diff
@@ -1,39 +1,39 @@
-{
```
maxime-poulain, line 1: We got EF power tools config files now both in the infra & the domain projects.
We should just stick with one config.

Feel free to decide whether the config files should be in the domain or infra project.
Usually I put that in the Domain project but it is just a personal preference

maxime-poulain (changes requested):

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/Training.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    public int TrainingId { get; set; }
+    public string TrainingTitle { get; set; }
+    public string TrainerFirstName { get; set; }
+    public string TrainerLastName { get; set; }
+    public List<TrainingTopic> Topics { get; set; } = new List<TrainingTopic>();
+    public List<string> TrainingLanguages { get; set; } = new List<string>();
```
maxime-poulain, line 12: I think you can safely remove that property, it is not an information we need to expose to the end user.

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList.cshtml
CelineMarique: They're already in a _Training_ folder inside of Pages
Do you mean you would separate the view model and the cshtml in different folders?

src/Showcase/src/Smart.FA.Catalog.Showcase.Infrastructure/DesignTimeContextFactory.cs
CelineMarique: No

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/Training.cs
CelineMarique: Ok, we can all decide to append view models like that

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/Training.cs
CelineMarique: In this context, but a training view model can be shared with a TrainingDetails page, it is the same model, and we do want to know the language of the training

maxime-poulain (changes requested):

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList.cshtml
```diff
@@ -0,0 +1,57 @@
...
+                    </div>
+                    <div class="training-list__trainer">
+                        @Html.DisplayFor(modelItem => training.TrainerFirstName) @Html.DisplayFor(modelItem => training.TrainerLastName)
+                    </div>
+                    <div class="training-list__sectors ">
+                        <img src="./icons/tag--blue.svg" class="o-svg-icon--small sectors-icon" />
+                        <!-- Show all training topics -->
+                        @foreach (var topic in training.Topics)
+                        {
+                            <a class="training-list__category-link" asp-page="TrainingViewModel/TrainingViewModelList" asp-route-cat="TopicName">@topic.Id</a>
```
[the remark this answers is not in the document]
CelineMarique: it's going to be `asp-route-id=@training.TrainingId`

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList.cshtml
Victor-van-Duijnen: @CelineMarique  It's really a matter of preference, but if we want to have more coherence between showcase and useradmin, I guess it would be better to have a folder per view where there would be your view model mappping logic (anyway we never know what next page we will need to be there in future releases). 

Although, you could still use the same view model between the pages!
We chose to go for separation of the view models for each view to avoid having one single "monolith" view model mapped for each view, but the same logic does not necessary apply to showcase since the data to be displayed is quiet different.

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList.cshtml.cs
```diff
@@ -0,0 +1,43 @@
...
+            .ToListAsync();
+
+        var trainingsByIds = trainingList.ToLookup(t => t.TrainingId);
+
+        foreach (var groupedTraining in trainingsByIds)
+        {
+            Trainings.Add(new TrainingViewModel()
+            {
+                TrainingId = groupedTraining.Key,
+                TrainingTitle = groupedTraining.First().TrainingTitle,
```
LievenPraet, line 33: If `groupedTraining` is empty, calling `First` will cause a crash.
Use `FirstOrDefault` instead and use the `?` operator.
Same remark for other instances where you use `First`.
```suggestion
                TrainingTitle = groupedTraining.FirstOrDefault()?.TrainingTitle,
```

LievenPraet (changes requested):

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList.cshtml.cs
CelineMarique: We do not call .First() at any time.

CelineMarique: We do not call .First() at any time for a training, We do it for the training status. Normally, there should be no trainings encoded without status. I can change it for .FirstOrDefault()

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList.cshtml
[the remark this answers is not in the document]
maxime-poulain: This is a hyperlink under a topic isn't it ? Shouldn't click on a link, let's say "marking", open all marketing trainings ?

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList.cshtml.cs
maxime-poulain: Nah at that point we now for a fact we have at least one record.
We grouped by the training id just before.

src/Showcase/src/Smart.FA.Catalog.Showcase.Domain/Models/TrainingList.cs
```diff
@@ -10,5 +10,6 @@ public partial class TrainingList
     public string TrainingLanguage { get; set; }
     public string TrainerFirstName { get; set; }
     public string TrainerLastName { get; set; }
```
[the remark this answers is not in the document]
maxime-poulain: Sadly we cannot anymore.
We went with SQL views.

In the first place, we did not want, unless impossible to do otherwise, sharing the domain layer of the UserAdmin Catalog part with the showcase website.

The UserAdmin domain comes with multiple dependencies that should not be a concern of a "public" website.

Therefore we considered those approaches :

- Making an actual API exposing the data that the showcase website needs.
- Using specific SQL views to scaffold actual DTO's for the showcase website.
- Scaffold database tables and use simple EF Core queries

Since 1) is not possible, we thought going with 2) would have been a good approach.
3) is actually the easiest as you don't depend of anything but still retarded as you got same definitions across two different projects.

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList.cshtml
maxime-poulain: >I was planning on sharing the view model with the upcoming TrainingDetails page

Usually you want to avoid re-using viewmodels accross pages, by definition a ViewModel exposes data to a given page.
The data that exposes "TrainingListViewModel" is for a training title training right ?
The ViewModel (or dto w/e) you gonna need for the detail of a training has more more fields such goal, methodology, etc.
The idea is also at some point to refactor the tile into a view component to make it reusable between the home page and the training list page.

However you pointed out something interesting: 
>There is no plan to have more views 'training related'

Why not create a base training view model that has common properties of every pages ?
If you don't want to have multiple files / folders, it is **OK** to have your page's viewmodel inside your XXX.cshtml.cs file.

CelineMarique: You can see in the last fixups that i followed your recommendations.

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList.cshtml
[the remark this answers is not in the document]
CelineMarique: I don't know yet the intention for those links. 
@joe-ubik , could you let us know if there will be some kind of page for a topic.
here I just followed the design introduced by Vio.

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Shared/_Layout.cshtml
```diff
@@ -23,7 +23,7 @@
                         <a class="header-nav__link" asp-area="" asp-page="/Index">Formateurs Associés</a>
                     </li>
                     <li class="header-nav__item">
-                        <a class="header-nav__link" asp-area="" asp-page="/Trainings">Formations</a>
+                        <a class="header-nav__link" asp-area="" asp-page="/Training/TrainingList/TrainingList">Formations</a>
```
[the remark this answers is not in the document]
CelineMarique: I don't want to use /Trainings. It was the default html inserted by Vio. I'd like to keep the route as it exist in the repository for more clarity. 
Is that a change request?

LievenPraet (approved):

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList.cshtml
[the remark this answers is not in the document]
maxime-poulain: Oki, can be adjusted anyway if required later on.

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Shared/_Layout.cshtml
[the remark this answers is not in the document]
maxime-poulain: Nope it's was not a change request. I just had the feeling "/Training/TrainingList/TrainingList" is a strange route.

maxime-poulain (approved):

PR #72, opened 2022-04-14
Add versioned user chart to home page and policies

Victor-van-Duijnen:
[The PR encapsulate the different tasks:](https://smartjira.atlassian.net/browse/CFA-242)

1. Add a user chart model with validity/expiration date
2. Add user chart to the trainer model
3. Create a policy on all common page 
_A trainer needs to have accepted at least one valid user chart to continue_
4. Create versioned pdf document in S3 storage and fetch pre-signed url (shared object in S3 terms)


CelineMarique (approved):

src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Commands/UploadUserChartCommand.cs
```diff
@@ -0,0 +1,43 @@
...
+        _context = context;
+    }
+
+    public Task<UploadUserChartResponse> Handle(UploadUserChartRequest request, CancellationToken cancellationToken)
+    {
+        UploadUserChartResponse response = new();
+        UserChart userChart = new(request.Title, request.Version, request.ValidityDate, request.ExpirationDate);
+        _context.UserCharts.Add(userChart);
+        response.UserChart = userChart;
+        response.SetSuccess();
```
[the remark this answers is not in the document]
Victor-van-Duijnen: Oops good point, Indeed it's not used yet, it will be a page reserved for root users so we can upload new user charts. But that doesn't mean it should not yet be correct obviously. I will change that!

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Services/BootStrapService.cs
```diff
@@ -28,13 +29,38 @@ public BootStrapService(ILogger<BootStrapService> logger, IServiceScopeFactory f
...
 
-        var filePath = Path.Combine(webRootPath, "default_image.jpg");
+        await UploadDefaultDocumentToS3Storage(filePath, fileName);
+    }
+
+    /// <inheritdoc />
+    public async Task AddDefaultUserChart(string webRootPath, string userChartName)
+    {
+        using var serviceScope = _factory.CreateScope();
+        var catalogContext = serviceScope.ServiceProvider.GetRequiredService<CatalogContext>();
```
LievenPraet, line 45: Can we directly inject the catalog context instead of going through the service scope?
Why the factory?

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Services/BootStrapService.cs
```diff
@@ -28,13 +29,38 @@ public BootStrapService(ILogger<BootStrapService> logger, IServiceScopeFactory f
...
+        var userChart = await catalogContext.UserCharts.OrderByDescending(userChart => userChart.CreatedAt).FirstOrDefaultAsync();
+        if (userChart is null)
+        {
+            throw new Exception("There are no user charts in DB");
```
LievenPraet, line 49: Try to throw a specific error instead of the generic `Exception`.
Could be an `InvalidOperationException` or you could find a more relevant exception if you don't agree.

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Services/BootStrapService.cs
```diff
@@ -28,13 +29,38 @@ public BootStrapService(ILogger<BootStrapService> logger, IServiceScopeFactory f
...
+
+        await UploadDefaultDocumentToS3Storage(filePath, userChartName);
+    }
+
+    private async Task UploadDefaultDocumentToS3Storage(string filePath, string fileName)
+    {
+        using var serviceScope = _factory.CreateScope();
+        var storageService = serviceScope.ServiceProvider.GetRequiredService<IS3StorageService>();
+
         var file = File.OpenRead(filePath);
```
LievenPraet, line 64: Could we use the `async` version of file read?
https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream.readasync?view=net-6.0

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Services/S3StorageService.cs
```diff
@@ -65,4 +65,17 @@ public async Task UploadAsync(Stream fileStream, string fileName, CancellationTo
...
+    /// Get the pre-signed url
+    /// </summary>
+    /// <param name="fileName">File to fetch</param>
+    /// <param name="expirationDate">Date of expiration of the generated token</param>
+    /// <returns>The Stream of the file</returns>
+    public Uri? GetPreSignedUrl(string fileName, DateTime expirationDate)
+    {
+        var absoluteUrl =_client.GetPreSignedURL(new GetPreSignedUrlRequest {Key = fileName, BucketName = _options.ImageBucketName, Expires = expirationDate}) ?? null;
+        var uri = absoluteUrl is null ? null : new Uri(absoluteUrl);
+        return uri;
```
LievenPraet, line 79: You can return right away without creating a var
```suggestion
return absoluteUrl is null ? null : new Uri(absoluteUrl);
```

src/UserAdmin/src/Smart.FA.Catalog.Web/Policies/Requirements/AtLeastOneValidUserChartApprovalRequirement.cs
```diff
@@ -0,0 +1,44 @@
...
+    public UserChartApprovalHandler(CatalogContext catalogContext, IHttpContextAccessor httpContextAccessor)
+    {
+        _catalogContext = catalogContext;
+        _httpContextAccessor = httpContextAccessor;
+    }
+
+    protected override Task HandleRequirementAsync(AuthorizationHandlerContext context, AtLeastOneValidUserChartApprovalRequirement requirement)
+    {
+        var currentDate = DateTime.UtcNow;
+        var identity = (CustomIdentity) context.User.Identity!;
```
LievenPraet, line 30: I guess `context.User.Identity` will never be null because we are in the context of an `AuthorizationHandler`?

src/UserAdmin/src/Smart.FA.Catalog.Web/Policies/Requirements/AtLeastOneValidUserChartApprovalRequirement.cs
```diff
@@ -0,0 +1,44 @@
...
+        var validUserChartsApprovals = _catalogContext.Trainers
+            .FirstOrDefault(trainer => trainer.Id == identity.Id)?
+            .Approvals
+            .Where(trainerApproval => (trainerApproval.UserChart.ValidityDate == null || trainerApproval.UserChart.ValidityDate!.Value.Ticks >= currentDate.Ticks) &&
+                                      (trainerApproval.UserChart.ExpirationDate == null || trainerApproval.UserChart.ExpirationDate!.Value.Ticks <= currentDate.Ticks));
+        if (validUserChartsApprovals is not null && validUserChartsApprovals.Any())
+        {
+            _httpContextAccessor.HttpContext!.Response.Cookies.Append("HasAcceptedUserChart", "true", new CookieOptions{MaxAge = TimeSpan.FromMinutes(1)});
+            context.Succeed(requirement);
+        }
```
LievenPraet, line 40: Do we need to handle the case where `validUserChartsApprovals` is null?

LievenPraet (changes requested):

src/UserAdmin/src/Smart.FA.Catalog.Web/Policies/Requirements/AtLeastOneValidUserChartApprovalRequirement.cs
Victor-van-Duijnen: Hmm the case where validUserChartApprovals is null is the case where there would not be any trainers of the id of the connected trainer. As of now, it's impossible for it to be null but if the ProxyHeaderMiddleware in the future is changed incorrectly, a null validUserChartApprovals should be treated as an empty one, meaning the user should not be authorized to proceed (since we cannot verify his identity and so not clear him to navigate cfa).

src/UserAdmin/src/Smart.FA.Catalog.Web/Policies/Requirements/AtLeastOneValidUserChartApprovalRequirement.cs
Victor-van-Duijnen: The ProxyHeadermiddleware at the beginning of each request make sure that there can be no unauthenticated call, this will indeed never be null (but indeed the authorization handler would not authorize it either but I am not counting on it)

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Migrations/20220413131809_Add_UserChart_Approval_To_Trainer.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            migrationBuilder.DropTable(
+                name: "TrainerApproval");
+
+            migrationBuilder.DropTable(
+                name: "UserChart");
+
+            migrationBuilder.DropColumn(
+                name: "ProfileImagePath",
+                table: "Trainer");
```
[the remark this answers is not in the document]
Victor-van-Duijnen: Good spot, I will remove that (it got added by mistake).

src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/GetLastUserChartUrlQuery.cs
```diff
@@ -0,0 +1,67 @@
...
+        _logger.LogInformation("Generated url for user chart pdf is {Url}", uriBuilder?.Uri.ToString() ?? string.Empty);
+
+        //TODO: We will need to figure out another workaround for Dev' and Staging deployment with devops since minio is dockerized and port mapping on host machine are not allowed
+        if (uriBuilder is not null && string.Equals(uriBuilder.Host, "host.docker.internal"))
+        {
+            uriBuilder.Scheme = "http";
+            uriBuilder.Host = "localhost";
+        }
+
+        return uriBuilder!.Uri;
```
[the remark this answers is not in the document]
Victor-van-Duijnen: Absolutely right, removing it

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Services/BootStrapService.cs
Victor-van-Duijnen: That method is NOT reading a file stream, it's just [creating a stream to be read](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.openread?view=net-6.0) (so there exist no asynchronous version of the method since it's actually just placing the cursor at the beginning of the file to be read). 
But from the result of that method (a Stream), you can absolutely read it asynchronously (as it is actually done bu the S3 service).

LievenPraet (approved):

src/UserAdmin/src/Smart.FA.Catalog.Web/Policies/Requirements/AtLeastOneValidUserChartRevisionApprovalRequirement.cs
```diff
@@ -0,0 +1,45 @@
...
+    public UserChartRevisionApprovalHandler(CatalogContext catalogContext, IHttpContextAccessor httpContextAccessor, IUserIdentity userIdentity)
+    {
+        _catalogContext = catalogContext;
+        _httpContextAccessor = httpContextAccessor;
+        _userIdentity = userIdentity;
+    }
+
+    protected override async Task HandleRequirementAsync(AuthorizationHandlerContext context, AtLeastOneValidUserChartRevisionApprovalRequirement requirement)
+    {
+        var currentDate = DateTime.UtcNow;
```
maxime-poulain, line 30:
```suggestion
        var currentDate = DateTime.UtcNow.Date;
```

maxime-poulain (changes requested):

src/UserAdmin/src/Smart.FA.Catalog.Web/Policies/Requirements/AtLeastOneValidUserChartRevisionApprovalRequirement.cs
maxime-poulain: This last little thing and good to be merged 👍

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/UserChart.cshtml.cs
```diff
@@ -0,0 +1,55 @@
...
+using Smart.FA.Catalog.Application.UseCases.Commands;
+using Smart.FA.Catalog.Application.UseCases.Queries;
+using Smart.FA.Catalog.Core.Services;
+
+namespace Smart.FA.Catalog.Web.Pages.Admin;
+
+public class UserChartModel : AdminPage
+{
+    public IUserIdentity UserIdentity { get; }
+    [BindProperty] public bool HasAcceptedUserChart { get; set; } = false;
```
maxime-poulain, line 12: Breakline

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/UserChart.cshtml.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    public ActionResult OnGet()
+    {
+        if (HttpContext.Request.Cookies.TryGetValue("AtLeastOneValidUserChartRevisionApproved", out _))
```
maxime-poulain, line 16: Since now it is a new different page, shouldn't the user just be redirect if he has access to services rather than checking in some cookie ?

If you made the query to say yes or no the user can access services, simply use it here 👍

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/UserChart.cshtml.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    public UserChartModel(IMediator mediator, IUserIdentity userIdentity) : base(mediator)
+    {
+        UserIdentity = userIdentity;
+    }
+
+    protected override SideMenuItem GetSideMenuItem()
+    {
+        // Doesn't matter we return MyTrainings or whatever, we don't see it anyway.
+        return SideMenuItem.MyTrainings;
```
maxime-poulain, line 32: You can make inhirit PageModel instead of AdminPage therefore.

maxime-poulain (changes requested):

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/UserChart.cshtml.cs
Victor-van-Duijnen: Good point, it's actually already what is is doing. it's trying to go straight to the home page but since they are not authorized, they are redirected to the user chart apporval page. I will remove the line!

maxime-poulain (approved):

PR #76, opened 2022-04-14
TrainerConfiguration: add constraint length on email and language

maxime-poulain:
This commit makes maximum length on email and default language respectively 254 and 2 characters long.
https://smartjira.atlassian.net/browse/CFA-325

### DIFF
--- a/src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Migrations/20220414183714_Add_Fixed_Length_On_Trainer_Language_And_Email.cs
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Migrations/20220414183714_Add_Fixed_Length_On_Trainer_Language_And_Email.cs
@@ -1,0 +1,51 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace Infrastructure.Migrations
+{
+    public partial class Add_Fixed_Length_On_Trainer_Language_And_Email : Migration
+    {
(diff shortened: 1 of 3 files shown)


src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Migrations/20220414183714_Add_Fixed_Length_On_Trainer_Language_And_Email.cs
```diff
@@ -0,0 +1,51 @@
...
+    public partial class Add_Fixed_Length_On_Trainer_Language_And_Email : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AlterColumn<string>(
+                name: "Email",
+                table: "Trainer",
+                type: "nvarchar(254)",
+                maxLength: 254,
+                nullable: true,
```
Victor-van-Duijnen, line 16: Just a thought, whenever updating his profile a trainer can absolutely not have a null email address (it's enforced in your validator). Should we not change the trainer constructor to force a profile with a valid mail address and set the email field to non-nullable?

maxime-poulain: Actually I am hesitating between
- totally remove the email part (revert stuff about it).
- remove the email from UI and validator but keep the column in the database as nullable.

That's something we need to decide actually x).

Victor-van-Duijnen (approved):

PR #94, opened 2022-04-28
Add support for super admin rights

maxime-poulain:
This introduces a simple mecanism to determine if a user is an admin or not .
It adds simply a Table holding some smart user ids that will be considered as admin.

### DIFF
--- a/src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/Authorization/HasSmartUserAdminRightsQuery.cs
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/Authorization/HasSmartUserAdminRightsQuery.cs
@@ -1,0 +1,42 @@
+using MediatR;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+using Smart.FA.Catalog.Core.Extensions;
+using Smart.FA.Catalog.Infrastructure.Persistence;
+
+namespace Smart.FA.Catalog.Application.UseCases.Queries.Authorization;
+
+public class HasSmartUserAdminRightsQuery : IRequest<bool>
+{
+    public string UserId { get; }
+
+    public HasSmartUserAdminRightsQuery(string userId)
+    {
+        UserId = userId;
+    }
+}
+
+public class HasSmartUserAdminRightsQueryHandler : IRequestHandler<HasSmartUserAdminRightsQuery, bool>
+{
+    private readonly ILogger<HasSmartUserAdminRightsQueryHandler> _logger;
+    private readonly CatalogContext _catalogContext;
+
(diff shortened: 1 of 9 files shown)


src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/Authorization/HasSmartUserAdminRightsQuery.cs
```diff
@@ -0,0 +1,45 @@
+using MediatR;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+using Smart.FA.Catalog.Core.Extensions;
+using Smart.FA.Catalog.Infrastructure.Persistence;
+
+namespace Smart.FA.Catalog.Application.UseCases.Queries.Authorization;
+
+public class HasSmartUserAdminRightsQuery : IRequest<bool>
```
LievenPraet, line 9:
```suggestion
public class SmartUserHasAdminRightsQuery : IRequest<bool>
```

LievenPraet (changes requested):

src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/Authorization/HasSmartUserAdminRightsQuery.cs
maxime-poulain: however I find it not really fancy as it starts with has.

src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Authorization/SuperAdmin.cs
```diff
@@ -0,0 +1,18 @@
+namespace Smart.FA.Catalog.Core.Domain.Authorization;
+
+public class SuperAdmin
```
chastelain, line 3: IMO I'd rather choose **SuperUser** instead of **SuperAdmin**, and the **Admin** should be an **Owner**

maxime-poulain: I ll go for super user , however renaming the admin part is not in the scope of this PR. But I ll already mentionned that to @LievenPraet , will be discussed later on.

src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/Authorization/HasSmartUserAdminRightsQuery.cs
maxime-poulain: Yeah I ll make it simple HasSuperUserRights, sound good ?

PR #100, opened 2022-04-30
Prevent application crashing upon startup if MinIO is not reachable

maxime-poulain:
Upon booting the application tries to add a default trainer image to MinIO.
If during the process any exception happens, the application will crash.
The application should be able to start instead of being inaccessible.

https://smartjira.atlassian.net/browse/CFA-363


Victor-van-Duijnen:
Then we need to go a step further and provide alternative methods for every action that would ask to store/retrieve a file.
So uploading/Fetching profile images should be disabled and the default user chart should be used (as far as the last actio, is concerned, it should already work, though). Or it will just crash constantly on those other actions.

maxime-poulain:
After discussion among us, it has been pointed out it would be preferred to have the startup failing if MinIO was down.
Closing this.

@Victor-van-Duijnen 
@thusc

PR #104, opened 2022-05-03
Staging and Production configurations

Victor-van-Duijnen:
The PR introduces settings for the staging and the production environement.
It includes:
- Kerberos settings for authentication in a linux container (with gss call)
- appsettings in preproduction and production
- Jeknins batch scripts


Victor-van-Duijnen:
Closing PR waiting for renaming of tables (big conflicts resolutions)

PR #107, opened 2022-05-03
Add smart field for super user

Victor-van-Duijnen:
This PR introduces a new IsGivenBySmart field accessible only by super-user in the create and update of a training and includes:
- Validation (only a super user can set the flag)
- A migration for the new field
- Domain model change

https://smartjira.atlassian.net/browse/CFA-334

### DIFF
--- a/src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Training/Training.cs
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Training/Training.cs
@@ -30,10 +30,10 @@ public class Training : SeedWork.Entity, IAggregateRoot
     public virtual IReadOnlyCollection<TrainingLocalizedDetails> Details => _details.AsReadOnly();
     public virtual IReadOnlyCollection<TrainingAttendance> Attendances => _attendances.AsReadOnly();
     public virtual IReadOnlyCollection<TrainingTopic> Topics => _topics.AsReadOnly();
-
     public int TrainerCreatorId { get; }
     public TrainingStatusType StatusType { get; private set; } = TrainingStatusType.Draft;
-     
+    public bool IsGivenBySmart { get; private set; }
+
     #endregion
 
     #region Constructors
@@ -64,6 +64,11 @@ public class Training : SeedWork.Entity, IAggregateRoot
     #endregion
 
     #region Public methods
+
+    public void MarkAsGivenBySmart(bool isFromSmart = true )
+    {
+        IsGivenBySmart = isFromSmart;
+    }
 
     public void SwitchVatExemptionTypes(IEnumerable<VatExemptionType>? vatExemptionTypes)
     {
--- a/src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/Trainings/Create/Index.cshtml
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/Trainings/Create/Index.cshtml
@@ -39,6 +39,10 @@
 
             <smart-panel header="@CatalogResources.FieldsThatWillAppearInThePublicCatalog">
                 <smart-form-group-input-text label="@CatalogResources.TrainingTitleCreationLabel *" asp-for="CreateTrainingViewModel.Title"></smart-form-group-input-text>
+                @if (Model.UserIdentity.IsSuperUser)
+                {
+                    <smart-checkbox label="@CatalogResources.IsTrainingGivenBySmart" asp-for="CreateTrainingViewModel.IsGivenBySmart"></smart-checkbox>
+                }
                 <smart-form-group-textarea rows="5" label="@CatalogResources.TrainingPresentation *" asp-for="CreateTrainingViewModel.Methodology" helper-text="@CatalogResources.Min30CharMax500Char"></smart-form-group-textarea>
                 <smart-form-group-textarea rows="5" label="@CatalogResources.TrainingGoal *" asp-for="CreateTrainingViewModel.Goal" helper-text="@CatalogResources.Min30CharMax500Char"></smart-form-group-textarea>
                 <smart-form-group-textarea rows="5" label="@CatalogResources.PracticalModalities *" asp-for="CreateTrainingViewModel.PracticalModalities" helper-text="@CatalogResources.Min30CharMax1000Char"></smart-form-group-textarea>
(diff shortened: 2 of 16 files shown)


src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/Trainings/Create/Index.cshtml
```diff
@@ -39,6 +39,10 @@
 
             <smart-panel header="@CatalogResources.FieldsThatWillAppearInThePublicCatalog">
                 <smart-form-group-input-text label="@CatalogResources.TrainingTitleCreationLabel *" asp-for="CreateTrainingViewModel.Title"></smart-form-group-input-text>
+                @if (Model.UserIdentity.IsSuperUser)
+                {
+                    <smart-checkbox label="@CatalogResources.IsTrainingGivenBySmart" asp-for="CreateTrainingViewModel.IsGivenBySmart"></smart-checkbox>
```
maxime-poulain, line 44: Download you put a <smart-validation-message /> but you did not here.

src/UserAdmin/src/Smart.FA.Catalog.Web/Validators/CreateTrainingRequestValidator.cs
```diff
@@ -10,24 +11,30 @@ namespace Smart.FA.Catalog.Web.Validators;
...
+            .WithMessage(CatalogResources.Max1000Characters);
+
+        RuleFor(viewModel => viewModel.PracticalModalities)
+            .MaximumLength(1000)
+            .WithMessage(CatalogResources.Max1000Characters);
 
-            RuleFor(viewModel => viewModel.PracticalModalities)
-                .MaximumLength(1000)
-                .WithMessage(CatalogResources.Max1000Characters);
+        When(_ => !userIdentity.IsSuperUser, () => RuleFor(viewModel => viewModel.IsGivenBySmart).NotEqual(true).WithMessage(CatalogResources.SuperUserPermissionToSetSmartTrainingType));
```
maxime-poulain, line 37:
```suggestion
        RuleFor(viewModel => viewModel.IsGivenBySmart)
            .Must(BeSuperUser)
            .When(GivenBySmartIsMarked)
            .WithMessage(CatalogResources.SuperUserPermissionToSetSmartTrainingType);      
```
It is always a good practice to use method groups when doing fluent chaining.

maxime-poulain, line 37:
```suggestion
        RuleFor(viewModel => viewModel.IsGivenBySmart)
            .Must(BeSuperUser)
            .When(GivenBySmartIsMarked)
            .WithMessage(CatalogResources.SuperUserPermissionToSetSmartTrainingType);      
```
It is always a good practice to use method groups when doing fluent chaining.

src/UserAdmin/src/Smart.FA.Catalog.Web/Validators/UpdateTrainingRequestValidator.cs
```diff
@@ -27,6 +28,8 @@ public UpdateTrainingViewModelValidator(IStringLocalizer<CatalogResources> local
             .MaximumLength(1000)
             .WithMessage(CatalogResources.Max1000Characters);
 
+        When(_ => !userIdentity.IsSuperUser, () => RuleFor(viewModel => viewModel.IsGivenBySmart).NotEqual(true).WithMessage(CatalogResources.SuperUserPermissionToSetSmartTrainingType));
```
maxime-poulain, line 31: Same remark than above.

maxime-poulain, line 31: Same remark than above.

src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Training/Training.cs
```diff
@@ -65,6 +65,11 @@ protected Training()
 
     #region Public methods
 
+    public void MarkAsGivenBySmart(bool isFromSmart = true )
```
LievenPraet, line 68: Should `isFromSmart` default to `true`?

Victor-van-Duijnen: I believe it does. From the name of the method it appears more clearly. If I write mytraining.MarkAsGivenBySmart(), I should have a training with the flag "IsGivenBySmart".
 It does not mean than a training will be created by default as "GivenBySmart" because the method is actually called by a parameter whose default value is false (but in the case nothing is specified in the signature of the method, seeing its name, it should default to true).

maxime-poulain: Hm going with Mark makes me suppose we got an UnMark, awell we short on time anyway x)

src/UserAdmin/src/Smart.FA.Catalog.Web/Validators/CreateTrainingRequestValidator.cs
```diff
@@ -10,24 +11,36 @@ namespace Smart.FA.Catalog.Web.Validators;
...
-                .MaximumLength(1000)
-                .WithMessage(CatalogResources.Max1000Characters);
+        RuleFor(viewModel => viewModel.PracticalModalities)
+            .MaximumLength(1000)
+            .WithMessage(CatalogResources.Max1000Characters);
 
-            RuleFor(viewModel => viewModel.PracticalModalities)
-                .MaximumLength(1000)
-                .WithMessage(CatalogResources.Max1000Characters);
+        RuleFor(viewModel => viewModel.IsGivenBySmart)
```
maxime-poulain, line 40: Not to be a pain if the ass, I have asked Lieven opinion.
```csharp
        RuleFor(viewModel => viewModel.IsGivenBySmart)
            .Must(BeSuperUser)
            .When(MarkedAsGivenBySmart)
            .WithMessage(CatalogResources.SuperUserPermissionToSetSmartTrainingType);
```

```csharp
private bool BeSuperUser(bool _) => _userIdentity.IsSuperUser;
```

maxime-poulain (changes requested):
just a little request, awell, if you dont agree fine by me to keep like this :)

src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/Training/Training.cs
Victor-van-Duijnen: Could you not just use MarkAsGivenBySmart(false)? I am open to suggestion for the naming, though

src/UserAdmin/src/Smart.FA.Catalog.Web/Validators/CreateTrainingRequestValidator.cs
maxime-poulain: You would have performance gain if there was an out parameter but in this case expect having a discard in the method signature, there's no difference at all.

Since the method is private and the bool is not used I guess it makes sense to have it discarded but on another end it's uglier to read so how's gonna be named the parameter, if you go this way it's entirely up to you :p.

LievenPraet (approved):

maxime-poulain (approved):

PR #108, opened 2022-05-04
Replace enumeration implementation with SmartEnum

maxime-poulain:
SmartEnum (https://github.com/ardalis/SmartEnum) is imported in replacement of the current implementation.
The current implementation is a bit limited and the API imposes strong typing on methods, which is not very convenient.


src/Shared/Smart.FA.Catalog.Shared/Domain/Enumerations/Common/Enumeration.cs
```diff
@@ -0,0 +1,348 @@
...
+            return LocalFromName(_fromNameIgnoreCase.Value);
+        else
+            return LocalFromName(_fromName.Value);
+
+        TEnum LocalFromName(Dictionary<string, TEnum> dictionary)
+        {
+            if (!dictionary.TryGetValue(name, out var result))
+            {
+                throw new EnumerationNotFoundException($"No {typeof(TEnum).Name} with Name \"{name}\" found.");
+            }
```
Victor-van-Duijnen, line 143: Missing breakline.

src/Shared/Smart.FA.Catalog.Shared/Domain/Enumerations/Common/Enumeration.cs
```diff
@@ -0,0 +1,348 @@
...
+    /// <seealso cref="Enumeration{TEnum,TValue}.TryFromValue(TValue, out TEnum)"/>
+    public static TEnum FromValue(TValue value)
+    {
+        if (value == null)
+            ArgumentNullException.ThrowIfNull(value);
+
+        if (!_fromValue.Value.TryGetValue(value, out var result))
+        {
+            throw new EnumerationNotFoundException($"No {typeof(TEnum).Name} with Value {value} found.");
+        }
```
Victor-van-Duijnen, line 212: Missing breakline.

Victor-van-Duijnen (approved):
Nice one, just added some missing breaklines my IDE picked up but as far as I am concerned it's good to go.

maxime-poulain:
> Nice one, just added some missing breaklines my IDE picked up but as far as I am concerned it's good to go.

I ported the source as it is :p.
You can also suggest break lines removals this is super easy to commit the suggestion.

PR #109, opened 2022-05-04
Add CreatedBy and LastModifiedBy properties to Entity base class

maxime-poulain:
This commits makes auditable the CreateBy, CreatedAt, LastModifiedBy and LastModifiedAt properties.
CatalogContext will ensure that those properties are correctly set upon saving.
The context will not update modified properties when inserting an entity nor will update created properties when updating an entity.
The CreatedBy and LastMofifiedBy properties are set accordingly to the Id of the current user.
CreatedAt and LastModifiedAt properties are set at DateTime.UtcNow if not ignored.


LievenPraet (approved):

src/UserAdmin/src/Smart.FA.Catalog.Web/Identity/UserIdentity.cs
```diff
@@ -7,26 +7,22 @@ namespace Smart.FA.Catalog.Web.Identity;
...
 {
+    private readonly IHttpContextAccessor _httpContextAccessor;
+
     /// <inheritdoc />
-    public int Id => Identity.Id;
+    public int Id => Identity?.Id ?? 0;
 
     /// <inheritdoc />
-    public CustomIdentity Identity { get; }
+    public CustomIdentity Identity => (_httpContextAccessor.HttpContext?.User.Identity as CustomIdentity)!;
```
Victor-van-Duijnen, line 16: It seems we should handled the case where Identity would not be castable to CustomIdentity

src/UserAdmin/src/Smart.FA.Catalog.Web/Identity/UserIdentity.cs
```diff
@@ -7,26 +7,22 @@ namespace Smart.FA.Catalog.Web.Identity;
...
 
     /// <inheritdoc />
-    public Trainer CurrentTrainer { get; }
+    public Trainer CurrentTrainer => Identity?.Trainer!;
```
Victor-van-Duijnen, line 19: Seeing the "Identity" method it will never return a null object
```suggestion
    public Trainer CurrentTrainer => Identity.Trainer;
```

src/UserAdmin/src/Smart.FA.Catalog.Web/Identity/UserIdentity.cs
```diff
@@ -7,26 +7,22 @@ namespace Smart.FA.Catalog.Web.Identity;
...
 
     public UserIdentity(IHttpContextAccessor httpContextAccessor)
     {
-        var identity = httpContextAccessor.HttpContext?.User.Identity as CustomIdentity ??
-                       throw new InvalidOperationException("Cannot retrieve identity of current user.");
-
-        Identity       = identity;
-        CurrentTrainer = identity.Trainer;
-
-        IsSuperUser = httpContextAccessor.HttpContext!.User.IsInRole("SuperUser");
```
Victor-van-Duijnen, line 30: Could we have a constant for the role "SuperUser" please? I know it's a temporary system anyway but we never know.

src/UserAdmin/src/Smart.FA.Catalog.Web/Identity/UserIdentity.cs
```diff
@@ -7,26 +7,22 @@ namespace Smart.FA.Catalog.Web.Identity;
 /// <inheritdoc />
 public class UserIdentity : IUserIdentity
 {
+    private readonly IHttpContextAccessor _httpContextAccessor;
+
     /// <inheritdoc />
-    public int Id => Identity.Id;
+    public int Id => Identity?.Id ?? 0;
```
Victor-van-Duijnen, line 13: It's pointless to use the coalesce operator in that case since identity will never be null (and Id neither).
```suggestion
    public int Id => Identity.Id;
```

src/UserAdmin/src/Smart.FA.Catalog.Web/Identity/UserIdentity.cs
Victor-van-Duijnen: If we are certain the HttpContext and that Identity cannot be null (since there are no fallback scenarios), then I believe that might be more correct:
```suggestion
    public CustomIdentity Identity => (_httpContextAccessor.HttpContext!.User.Identity as CustomIdentity)!;
```

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Persistence/Extensions/AuditableEntityExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+namespace Smart.FA.Catalog.Infrastructure.Persistence.Extensions;
+
+public static class EntryEntityAuditableDateUpdate
+{
+    /// <summary>
+    /// Updates a collection of <see cref="Entity" /> to keep their data auditable.
+    /// </summary>
+    /// <param name="entityEntries">A collection of <see cref="EntityEntry{TEntity}"/>.</param>
+    /// <param name="userId">The id of the user doing the current operation.</param>
+    public static void UpdateAutitableEntitiesData(this IEnumerable<EntityEntry<Entity>> entityEntries, int userId)
```
Victor-van-Duijnen, line 14:
```suggestion
    public static void UpdateAuditableEntitiesData(this IEnumerable<EntityEntry<Entity>> entityEntries, int userId)
```

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Persistence/Extensions/AuditableEntityExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        foreach (var entityEntry in entityEntries)
+        {
+            entityEntry.UpdateEntityAuditableData(userId);
+        }
+    }
+
+    private static void UpdateEntityAuditableData(this EntityEntry<Entity> entityEntry, int userId)
```
Victor-van-Duijnen, line 22: To keep a coherent naming:
```suggestion
    private static void UpdateAuditableEntityData(this EntityEntry<Entity> entityEntry, int userId)
```

src/UserAdmin/src/Smart.FA.Catalog.Web/Identity/UserIdentity.cs
maxime-poulain: It is already the case : when you seed the data you don't have any identity. But that's a particular case and I don't want to expose to the other developer a nullable api because of one exceptional case.

src/UserAdmin/src/Smart.FA.Catalog.Web/Identity/UserIdentity.cs
maxime-poulain: When seeding it is equal to zero.

src/UserAdmin/src/Smart.FA.Catalog.Web/Identity/UserIdentity.cs
maxime-poulain: I thought of it and was pretty sure you ll ask for.
But I don't really want to create a class for something we know will be in the next sprints after the go in production.

Victor-van-Duijnen (approved):

PR #119, opened 2022-05-09
Disable event on key enter pressed on form in training creation and edition pages

maxime-poulain:
Just disable submitting the form when pressing enter on the create and edit page of trainings.
Hitting ENTER submit as a draft but we don't want necessarily to have that behavior.


joe-coop (approved):

maxime-poulain:
Merging, can be reverted in any case.

PR #123, opened 2022-05-10
Rewrite paging query for sql server 2008 support

Victor-van-Duijnen:
This PR introduces a hotfix for the paging system.
FETCH/NEXT query commands are unsupported in sql Server 2008 and had to be replaced by a ROW_Number instruction for backward compatibility.  There used to be a way to specifiy an options for forcing ef core to use the older SQL Server 2008 syntax but its support was dropped ([read ref](https://docs.microsoft.com/en-us/ef/core/what-is-new/ef-core-3.x/breaking-changes#urn)).


Victor-van-Duijnen:
We found a better way than to rewrite manually each queries using paging

PR #128, opened 2022-05-10
Standarize paging super user

Victor-van-Duijnen:
In previous PR, a change of the paging of trainings broke the paging of the super user but was left untested.
Here is a hotfix to make an unified system for the paged list.

### DIFF
--- a/src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Extensions/PagedListExtension.cs
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Extensions/PagedListExtension.cs
@@ -5,7 +5,7 @@ namespace Smart.FA.Catalog.Infrastructure.Extensions;
 
 public static class PagedListExtension
 {
-    public static async Task<PagedList<T>> PaginateAsync<T>(this IQueryable<T> source, PageItem pageItem)
+    public static async Task<PagedList<T>> PaginateAsync<T>(this IQueryable<T> source, PageItem pageItem, CancellationToken cancellationToken)
     {
         var count = await source.CountAsync();
         var items = await source.Skip(pageItem.Offset).Take(pageItem.PageSize).ToListAsync();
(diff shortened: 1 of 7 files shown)


src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Extensions/PagedListExtension.cs
```diff
@@ -5,7 +5,7 @@ namespace Smart.FA.Catalog.Infrastructure.Extensions;
 
 public static class PagedListExtension
 {
-    public static async Task<PagedList<T>> PaginateAsync<T>(this IQueryable<T> source, PageItem pageItem)
+    public static async Task<PagedList<T>> PaginateAsync<T>(this IQueryable<T> source, PageItem pageItem, CancellationToken cancellationToken)
```
Khaon, line 8: You pass the token but you don"t pass it down to the ToListAsync method.
Make it optional i.e. CancellationToken cts = default

I made a misstake but every cancellation tokens should be optional.

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/SuperUser/Trainings/List/SuperUserTrainingList.cshtml
```diff
@@ -111,7 +111,7 @@
                 }
             </div>
 
-            <div>Page @(Model.Trainings.PageIndex.ToString())/@(Model.Trainings.TotalPages.ToString())</div>
+            <div>Page @(Model.Trainings.PageIndex + 1)/@(Model.Trainings.TotalPages.ToString())</div>
```
Khaon, line 114: You should switch to the CurrentPage of yours instead of +1 on PageIndex.
PageIndex in the late implementation PaginatedList you removed is the equivalent of the CurrentPage.

Khaon (changes requested):

maxime-poulain (approved):

Khaon (approved):

PR #131, opened 2022-05-10
Add special authentication options for using fake headers

Victor-van-Duijnen:
This PR simply add an option for allowing/forbidding fake headers for testing with/out nginx.

### DIFF
--- a/src/UserAdmin/src/Smart.FA.Catalog.Web/appsettings.json
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Web/appsettings.json
@@ -45,7 +45,7 @@
   "SuperUser": {
       "NumberOfTrainingsPerPage": 20
   },
-  "AuthenticationOptions": {
+  "Authentication": {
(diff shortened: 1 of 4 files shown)


src/UserAdmin/src/Smart.FA.Catalog.Web/appsettings.json
```diff
@@ -44,5 +44,8 @@
   },
   "SuperUser": {
       "NumberOfTrainingsPerPage": 20
+  },
+  "AuthenticationOptions": {
```
maxime-poulain, line 48: Could you just remove the "Options" suffix in the json part ?

maxime-poulain: Just force merge after doing or not that part.

Victor-van-Duijnen: Ok I did

PR #139, opened 2022-05-11
Prefix path for nginx redirection

Victor-van-Duijnen:
This PR introduces necessary changes for a proper Nginx redirection.
Notably to add systematically a keyword (here cfa) that can be picked up by both Account and CFA.

- It's not ideal but the keyword "cfa" was preprended manually to all pages.
- It also redirects to account aspx page on the home logo and the logout. Note there are no fallback routes as of now, either.


Victor-van-Duijnen:
We found a better way to do that! PR incoming.

PR #141, opened 2022-05-12
Replace pre-signed image by public url

Victor-van-Duijnen:
This PR replaces pre-signed image by public url. The image's path are hashed but publically accessible, limiting the number of calls to the application and its performance.


src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Helpers/MinIoLinkGenerator.cs
```diff
@@ -0,0 +1,52 @@
+﻿using HashidsNet;
+using Microsoft.Extensions.Options;
+using Smart.FA.Catalog.Infrastructure.Services.Options;
+
+namespace Smart.FA.Catalog.Infrastructure.Helpers;
+
+public class MinIoLinkGenerator : IMinIoLinkGenerator
```
LievenPraet, line 7: Service might be a better name to stay consistent in our naming
```suggestion
public class MinIoLinkService : IMinIoLinkService
```

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Helpers/IMinIoLinkGenerator.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace Smart.FA.Catalog.Infrastructure.Helpers;
+
+public interface IMinIoLinkGenerator
```
LievenPraet, line 3:
```suggestion
public interface IMinIoLinkService
```

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Shared/_AdminLayout.cshtml
```diff
@@ -53,8 +56,7 @@
                     <div class="c-toolbar__item">
                         <a class="c-user-navigation" href="#" data-menu="userMenu">
                             <div class="c-avatar c-avatar--img">
-                                <img src="/admin/myprofile?handler=LoadImage" alt="avatar">
-
+                                <img src="@(UserIdentity.CurrentTrainer.ProfileImagePath is null ? MinIoLinkGenerator.GetFullTrainerProfilePictureUrl(MinIoLinkGenerator.GetDefaultFullProfilePictureImageUrl()) : MinIoLinkGenerator.GetFullTrainerProfilePictureUrl(UserIdentity.CurrentTrainer.ProfileImagePath))" alt="avatar">
```
LievenPraet, line 59: We could create a local function to make the code a bit more readable

LievenPraet (changes requested):

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Helpers/MinIoLinkGenerator.cs
maxime-poulain: I suggested Victor that name, I think it is good as it makes it clear what it does.
An example:
https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.identity.passwordhasher-1?view=aspnetcore-6.0

However MinIOService we have does a bunch of different things it makes sense have suffix.

You will also most likely name a service that provides IConnection IConnectionProvider.

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Shared/_AdminLayout.cshtml
maxime-poulain: Two solutions here:
MinIoLinkGenerator.GetCurrentTrainerProfilePictureUrl() that uses the Current User Identity service to retrieves dynamically the URL with a fallback of the defualt one

OR

MinIoLinkGenerator.GetFullTrainerProfilePictureUrl(string path)=> if null return default;

But what's it is sure is that that line has too much logic.

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Helpers/IMinIoLinkGenerator.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    /// <summary>
+    /// Create the relative URL of the profile picture of the trainer out of his id
+    /// </summary>
+    /// <param name="trainerId">The id of the trainer</param>
+    /// <returns>The relative url to the profile picture of the trainer</returns>
```
maxime-poulain, line 9: It's more of a path nope?

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Helpers/MinIoLinkGenerator.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        _storageOptions = storageOptions.Value;
+    }
+
+    /// <inheritdoc />
+    public string CreateTrainerProfilePictureUrl(int trainerId, string imageFormat)
+    {
+        // Generate unique salt value to avoid decoding
+        var hashIds = new Hashids(MinioLinkDefaultSaltValues.TrainerProfilePicture, 8);
+        return Path.Combine("trainers/", "profile-image/", $"{hashIds.Encode(trainerId)}{imageFormat}");
```
maxime-poulain, line 21: If you don't like putting that in the configuration files (fine by me) you should, i think, extract them in private constant at the top of the class

maxime-poulain (changes requested):

maxime-poulain (approved):

LievenPraet (approved):

PR #147, opened 2022-05-13
LoggingPipelineBehavior: don't log Trainer nor Training domain models that transit through MediatR

maxime-poulain:
Due to circular references between properties of entities like Trainer and Training (they are aggregate/entities after all) serialization 
 produces very large JSON because of the lazy loading enable in EF Core.

Serializing an object make the serializer accessing every public properties; this trigger lazy evaluation of each unloaded ones which can result in unnecessary and heavy SQL statements.

Commit 9188b484c36031536ac373fbd0691b381962e23a serializes request that results in an exception, I was to avoid the issues described above.

This commit adds five JsonConverters on entities that are sent between web and BL to convert them to their primary key value rather the whole object value.

This should be considered as a dirty fix and DTOs should be considered over entities; or domain models that are not entities under the hood.


chastelain (approved):

maxime-poulain:
thx

PR #152, opened 2022-05-13
Add a 'cfa' base path to user admin URL so the website works with nginx

maxime-poulain:
In production the redirection on cfa will add a /cfa base path this add the necessary things to make the site works.

PR #157, opened 2022-05-13
Fix redirection to home page from the 404 page

maxime-poulain:
This commit changes the path to the home page from '/' to '/cfa' as we use now /cfa for the base path.


Victor-van-Duijnen (approved):

maxime-poulain:
Closing to be integrated in https://github.com/smartcoop/smart-cfa/pull/152

PR #162, opened 2022-05-16
Reload edit profile page after a successful edition

maxime-poulain:
Currently the page is re-rendered  to display a successful message, however doing so won't reload the newly uploaded image (if any).
This commit ensures that the page reloads itself and therefore refresh the trainer's profile image.

https://smartjira.atlassian.net/browse/CFA-422


Victor-van-Duijnen (approved):
Nice, indeed noticed that too but didn't make the change yet, thanks

PR #175, opened 2022-05-16
Change wording for Signing out and Routing in lowercase

Victor-van-Duijnen:
This PR:

1. Sets all routes to lower-case
2. Change wording "Se connecter" to "Déconnexion" to be coherent with Account

### DIFF
--- a/src/UserAdmin/src/Smart.FA.Catalog.Web/Extensions/MvcBuilderExtensions.cs
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Web/Extensions/MvcBuilderExtensions.cs
@@ -7,6 +7,13 @@ public static class MvcBuilderExtensions
 {
     public static IMvcBuilder ConfigureRazorPagesOptions(this IMvcBuilder mvcBuilder)
     {
+        // Sets all routes and query strings in lowercase
+        mvcBuilder.Services.Configure<RouteOptions>(options =>
+        {
+            options.LowercaseUrls = true;
+            options.LowercaseQueryStrings = true;
+        });
+
         mvcBuilder.AddRazorPagesOptions(options =>
         {
             options.Conventions.AddPageConventionsAuthorization();
--- a/src/UserAdmin/src/Smart.FA.Catalog.Web/Extensions/MvcBuilderExtensions.cs
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Web/Extensions/MvcBuilderExtensions.cs
@@ -7,13 +7,6 @@ public static class MvcBuilderExtensions
 {
     public static IMvcBuilder ConfigureRazorPagesOptions(this IMvcBuilder mvcBuilder)
     {
-        // Sets all routes and query strings in lowercase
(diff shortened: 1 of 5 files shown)


src/UserAdmin/src/Smart.FA.Catalog.Web/Extensions/MvcBuilderExtensions.cs
```diff
@@ -7,6 +7,13 @@ public static class MvcBuilderExtensions
 {
     public static IMvcBuilder ConfigureRazorPagesOptions(this IMvcBuilder mvcBuilder)
     {
+        // Sets all routes and query strings in lowercase
+        mvcBuilder.Services.Configure<RouteOptions>(options =>
+        {
+            options.LowercaseUrls = true;
+            options.LowercaseQueryStrings = true;
```
maxime-poulain, line 14: If you got a query string used for some search that is case sensitive how does that gonna behave with this setting to true?

Victor-van-Duijnen: It only affect the key of the query parameters, not the values. So it will work fine!

maxime-poulain (approved):

maxime-poulain (changes requested):
C:\Users\pam\Pictures\Screenpresso\video_08.mp4

The querystring's values are also set to lower case.

src/UserAdmin/src/Smart.FA.Catalog.Web/Extensions/MvcBuilderExtensions.cs
```diff
@@ -7,6 +7,13 @@ public static class MvcBuilderExtensions
 {
     public static IMvcBuilder ConfigureRazorPagesOptions(this IMvcBuilder mvcBuilder)
     {
+        // Sets all routes and query strings in lowercase
+        mvcBuilder.Services.Configure<RouteOptions>(options =>
```
maxime-poulain, line 11: I guess we should use the AddRouting extensionss method:
[    public static IServiceCollection AddRouting(](https://github.com/dotnet/aspnetcore/blob/c85baf8db0c72ae8e68643029d514b2e737c9fae/src/Http/Routing/src/DependencyInjection/RoutingServiceCollectionExtensions.cs#L113) rather then that one.

Note both of them are extensions on IServiceCollection not MvcBuilder.

maxime-poulain (changes requested):

src/UserAdmin/src/Smart.FA.Catalog.Web/Extensions/MvcBuilderExtensions.cs
Victor-van-Duijnen: Ah no that's not a query string parameter, a query string parameter has the form "?parameterstring=xxx". On your example it's part of the route unfortunately so indeed that's set to lowercase

maxime-poulain (approved):

PR #180, opened 2022-05-18
UserAdmin: SuperUserTrainingList: add support for searching trainings by training creator name

maxime-poulain:
Now super users can filter when searching by a trainer's name.

https://smartjira.atlassian.net/browse/CFA-426

### DIFF
--- a/src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/GetTrainingsByCriteriaQuery.cs
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/GetTrainingsByCriteriaQuery.cs
@@ -40,7 +40,7 @@ public class GetTrainingsByCriteriaQueryHandler : IRequestHandler<GetTrainingsBy
     public async Task<PagedList<Training>> Handle(GetTrainingsByCriteriaQuery request, CancellationToken cancellationToken)
     {
         // Includes necessary navigation properties.
-        var query = BaseQuery();
+        var query = BuildBaseQuery();
 
         // Filter by status.
         if (request.Status is not null)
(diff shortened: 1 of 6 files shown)


src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/GetTrainingsByCriteriaQuery.cs
```diff
@@ -61,8 +63,26 @@ public async Task<PagedList<Training>> Handle(GetTrainingsByCriteriaQuery reques
...
+            query = query.Where(training => training.TrainerAssignments.Any(assignment => assignment.Trainer.Name.FirstName.Contains(request.TrainerName) || assignment.Trainer.Name.LastName.Contains(request.TrainerName)));
+        }
+
+        // Finally apply pagination.
         var trainings = await query.PaginateAsync(new PageItem(request.PageNumber, request.PageSize), cancellationToken);
 
         return trainings;
     }
+
+    private IQueryable<Training> BaseQuery()
```
LievenPraet, line 78:
```suggestion
    private IQueryable<Training> GetBaseQuery()
```

LievenPraet (changes requested):

src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/GetTrainingsByCriteriaQuery.cs
maxime-poulain: I tend to put a "Get" prefix on method that gonna fetch one way or another some kind of resource.
"GetTrainerById" -> from db
"GetCountryList" => from an API or so.
To me, it makes it clear that we are fetching something and there's an underlying IO operation under the hood.

I might turn that into a public get property however ?

LievenPraet: To me, a function should represent an action: get, insert, delete, convert, send...
IN this case, a property might be a better choice here

LievenPraet (changes requested):
See comments above

Victor-van-Duijnen (approved):

PR #184, opened 2022-05-20
Reorganize solution structure

Victor-van-Duijnen:
This PR rename projects in useradmin and restructure solution items and scripts.


maxime-poulain:
Let"s not rename the namespaces for the time being.
It is not because there is a virtual folder called XXX that the namespace must include the name.

Furthermore, there's gonna be also some thought how to move the code into Opportunity (Smart.Coop) repository.

Parts of the current base need to be moved into common projects since the code will be part of bigger plans.
However moving config files into separate folders instead of having them at the root is something that can definitely be kept.

Victor-van-Duijnen:
Closed to introduce PR #192  without renaming of projects

PR #192, opened 2022-05-24
Organize solution items and dockerfiles (no renaming)

Victor-van-Duijnen:
This PR structure solution items and configuration for dockerfiles without the renaming of projects like in PR #184


LievenPraet (approved):

src/UserAdmin/src/Smart.FA.Catalog.Web/appsettings.Production.json
```diff
@@ -43,6 +43,6 @@
     "Showcase": "https://learning.smart.coop"
   },
   "Authentication": {
-    "UseFakeHeaders": false
```
maxime-poulain, line 46: Dunno if it is intended or not ?

maxime-poulain (changes requested):
someting to check with the production config.

src/UserAdmin/src/Smart.FA.Catalog.Web/appsettings.Production.json
Victor-van-Duijnen: Yes it's intended, although indeed temporary because Account is not yet in Production and it will allow us to test that environment  before it is (I know it's forbidden to "test" the production environement but I just mean testing in the sense of checking that the apps in the context of the infra interact properly together).

PR #194, opened 2022-05-25
Improve the display of superadmin page

joe-coop:
- Correct and add some localization keys
- Improve display screen


joe-coop:
not ready

PR #212, opened 2022-06-03
Feature/showcase add views migrations

maxime-poulain:


src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Program.cs
```diff
@@ -70,6 +70,11 @@
 
 app.MapRazorPages();
 
+if (!app.Environment.IsProduction() && !app.Environment.IsStaging())
```
maxime-poulain, line 73: is thje check on staging worth having ?

Victor-van-Duijnen: Hmm I think you should add a check on the "PreProduction" env'. 
You can remove the staging check. I intended to add showcase to the staging docker compose (unless we decide -like discussed yesterday- to move it to another repo) but anyway,  migrations are allowed since the db server is containerized in Staging.

maxime-poulain, line 73:
```suggestion
if (!app.Environment.IsProduction() && !app.Environment.IsEnvironment("PreProduction"))
```

maxime-poulain: @Victor-van-Duijnen

Victor-van-Duijnen (approved):

PR #213, opened 2022-06-03
Add Soft delete to entities

Victor-van-Duijnen:
[This PR introduces soft delete to entities, it allows:](https://smartjira.atlassian.net/browse/CFA-329)

- Entities to be destroyed (hard delete) with the Destroy() method
- Entities to be soft deleted with the Delete() method
=> It follows that I have used the flag IsDeleted to flag an entity as soft deleted

- It adds a global query filter on class deriving from the Entity class in user admin 
- It filters out soft deleted entities in sql views in showcase


src/Showcase/src/Smart.FA.Catalog.Showcase.Infrastructure/Data/Migrations/20220603102434_Filter_On_SoftDelete.cs
```diff
@@ -0,0 +1,241 @@
...
+                FROM [Cfa].Trainer trainer
+                    LEFT JOIN Cfa.TrainerSocialNetwork socialNetwork
+                    ON trainer.Id = socialNetwork.TrainerId
+                    INNER JOIN Cfa.TrainerAssignment assignment
+                    ON assignment.TrainerId = trainer.Id
+                    INNER JOIn Cfa.Training training
+                    ON training.Id = assignment.TrainingId
+                    INNER JOIN Cfa.TrainingStatusType trainingStatusType
+                    ON trainingStatusType.Id = training.TrainingStatusTypeId
+                WHERE TrainingStatusType.Name = 'Published'
```
LievenPraet, line 151: Did we forget to check to `IsDeleted` flag here?

src/Showcase/src/Smart.FA.Catalog.Showcase.Infrastructure/Data/Migrations/20220603102434_Filter_On_SoftDelete.cs
```diff
@@ -0,0 +1,241 @@
...
+                    ,trainer.ProfileImagePath
+                    ,count(assignment.TrainingId) As TrainingCount
+                FROM Cfa.Trainer trainer
+                    INNER JOIN Cfa.TrainerAssignment assignment
+                    ON assignment.TrainerId = trainer.Id
+                    INNER JOIN Cfa.Training training ON
+                    assignment.TrainingId = training.Id
+                    INNER JOIN Cfa.TrainingStatusType trainingStatusType
+                    ON trainingStatusType.Id = training.TrainingStatusTypeId
+                WHERE TrainingStatusType.Name = 'Published'
```
LievenPraet, line 180: Same remark as above

src/Showcase/src/Smart.FA.Catalog.Showcase.Infrastructure/Data/Migrations/20220603102434_Filter_On_SoftDelete.cs
```diff
@@ -0,0 +1,241 @@
...
+                    ON training.Id = details.TrainingId
+                    INNER JOIN [Cfa].[TrainerAssignment] trainingTrainer
+                    ON training.Id = trainingTrainer.TrainingId
+                    INNER JOIN [Cfa].[Trainer] trainer
+                    ON trainingTrainer.TrainerId = trainer.Id
+                    INNER JOIN [Cfa].[TrainingTopic] topic
+                    ON training.Id = topic.TrainingId
+                    INNER JOIN Cfa.TrainingStatusType trainingStatusType
+                    ON trainingStatusType.Id = training.TrainingStatusTypeId
+                WHERE TrainingStatusType.Name = 'Published'");
```
LievenPraet, line 211: Same remark as above

src/Showcase/src/Smart.FA.Catalog.Showcase.Infrastructure/Data/Migrations/20220603102434_Filter_On_SoftDelete.cs
```diff
@@ -0,0 +1,241 @@
...
+                    ON training.Id = detail.TrainingId
+                    INNER JOIN [Cfa].[TrainingTopic] category
+                    ON training.Id = category.TrainingId
+                    INNER JOIN [Cfa].[TrainerAssignment] trainingTrainer
+                    ON training.Id = trainingTrainer.TrainingId
+                    INNER JOIN [Cfa].[Trainer] trainer
+                    ON trainingTrainer.TrainerId = trainer.Id
+                    INNER JOIN Cfa.TrainingStatusType trainingStatusType
+                    ON trainingStatusType.Id = training.TrainingStatusTypeId
+                WHERE TrainingStatusType.Name = 'Published'");
```
LievenPraet, line 238: Same remark as above

src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/IEFSoftDelete.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Smart.FA.Catalog.Core.Domain;
+
+/// <summary>
+/// Add Soft Delete field to an entity
+/// </summary>
+public interface IEFSoftDelete
```
LievenPraet, line 6: I would suggest adding the full name of EF here for clarity.
```suggestion
public interface IEntityFrameworkSoftDelete
```

LievenPraet: On second thought, shouldn't it be called `IEntitySoftDelete`?

src/UserAdmin/src/Smart.FA.Catalog.Core/SeedWork/Entity.cs
```diff
@@ -45,6 +50,12 @@ public void ClearDomainEvents()
         _domainEvents.Clear();
     }
 
+    public void Delete() => IsDeleted = true;
+
+    public void Recover() => IsDeleted = false;
+
+    public void Destroy() => IsDestroyed = true;
```
LievenPraet, line 57: Wouldn't destroying the entity mean deleting it entirely from the database?
The functions `Delete` and `Destroy` now basically do the same job and are redundant.

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Extensions/SoftDeleteQueryExtension.cs
```diff
@@ -0,0 +1,47 @@
...
+                ApplySoftDeleteQueryFilter(entityType);
+            }
+        }
+    }
+
+    private static void ApplySoftDeleteQueryFilter(IMutableEntityType entityType)
+    {
+        // Get and call the static Lambda expression which filters soft deletable entities
+        var methodToCall = typeof(SoftDeleteQueryExtension).GetMethod(nameof(GetSoftDeleteFilter), BindingFlags.NonPublic | BindingFlags.Static)!.MakeGenericMethod(entityType.ClrType);
+        var filter = methodToCall.Invoke(null, Array.Empty<object>());
```
LievenPraet, line 30: Do we need to use reflection here?
It screws up IntelliSense and complicates the code.
Is there another way we could do this?

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Persistence/CatalogContext.cs
```diff
@@ -79,4 +81,18 @@ private Task PublishEntityDomainEventsAsync()
 
         return _eventPublisher.PublishEntitiesEventsAsync(entitiesWithEvents);
     }
+
+    private void SetSoftDeleteToEntities()
+    {
+        var entityListToSoftDelete = ChangeTracker.Entries<IEFSoftDelete>().Where(entry => entry.State == EntityState.Deleted);
+        foreach (var entity in entityListToSoftDelete)
+        {
+            // The IsDestroyed flag allow to force hard delete to the entity
```
LievenPraet, line 90: Why implement this on `ModelCreating`?
Can't we just delete the entity at the time we would set the `IsDestroyed` flag to true?

LievenPraet (changes requested):

maxime-poulain:
The concept feels a bit akward to me; having the entities responsible to determine their delete state feels strange.

To me, deleting / softdeleting should be completly the responsablity of your DAL (db context, uow, w/e).

Did you inspire on some resource you might point to have a look at ?

Victor-van-Duijnen:
In all of the references I have seen, they don't use methods but they have a public property IsDeleted on their object, which is the same as to expose a Delete methods because an object could still be responsible for its own destruction (myObject.IsDeleted = true, but then I would rather go with a declarative style like for all of our rich domain models than a public property) refs: https://www.thereformedprogrammer.net/ef-core-in-depth-soft-deleting-data-with-global-query-filters ; https://spin.atomicobject.com/2019/01/29/entity-framework-core-soft-delete ; https://medium.com/@uslperera/soft-delete-with-ef-core-c677bff73ef7... Conceptually it doesn't bother me personally because an entity is already modifying all of its properties and a soft delete is just another state of an object. Maybe we can find a compromise if the Destroy(hard delete) method is moved to an overload of specific db sets...

src/UserAdmin/src/Smart.FA.Catalog.Core/Domain/IEFSoftDelete.cs
Victor-van-Duijnen: Yes I am not too fond of my name either retrospectively. Why not IEntitySoftDelete indeed but what do you say about simplifying it to `ISoftDelete `?

LievenPraet: Yes, I was think ing about that after my remark as well :D
Make it `ISoftDelete`

maxime-poulain:
There is also an important case to be taken in account.
When you soft delete you save the entity.
However if someone changes an object than just call softdelete, the other column will be updated in the database.
I guess that's something we don't want.

src/UserAdmin/src/Smart.FA.Catalog.Core/SeedWork/Entity.cs
Victor-van-Duijnen: It was decided that we will remove the Destroy method to keep YAGNI principle :)

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Extensions/SoftDeleteQueryExtension.cs
```diff
@@ -0,0 +1,47 @@
...
+using Smart.FA.Catalog.Core.Domain;
+
+namespace Smart.FA.Catalog.Infrastructure.Extensions;
+
+/// <summary>
+/// Add global query filter for soft deletable class implementing the <see cref="ISoftDeletable"/> interface
+/// </summary>
+public static class SoftDeleteQueryExtension
+{
+    public static void AddSoftDeleteQueryFilter(this ModelBuilder modelBuilder)
```
maxime-poulain, line 14: You could do something better like this:
```csharp
public static void SetFilterQueryOnEntityDerivedFrom<T>(this ModelBuilder modelBuilder, Expression<Func<T, bool>> query)
{
    foreach (var mutableEntityType in modelBuilder.Model.GetEntityTypes().Where(entity => entity.GetType().IsAssignableTo(typeof(T))))
    {
        // At runtime the parameter type is expected to be `mutableEntityType.ClrType` but query's type if T.
        // So we need to revisit the expression tree to change the parameter type to the actual entity type.
        var actualEntityTypeParameter = Expression.Parameter(mutableEntityType.ClrType);
        var body = ReplacingExpressionVisitor.Replace(query.Parameters.First(), actualEntityTypeParameter, query.Body);
        var lambdaExpression = Expression.Lambda(body, actualEntityTypeParameter);

        mutableEntityType.SetQueryFilter(lambdaExpression);
    }
}

// usage :
modelBuilder.SetFilterQueryOnEntityDerivedFrom<IEntity>(e => e.Id < 5);
```

Victor-van-Duijnen:
Following the COPBACK it was decided we might close this PR since it was found the complexity was not worth it.
Before closing it definitively I would just like to have your comment @Geraldineb1 @pierrewatelet @BomUbik @JOABAUDENS so it doesn't look like a whim on my part.

maxime-poulain:
You an close it but we can go back at some point get the impl for the global query filtering, etc as the implementation is already there.

I doubt that among all the person you tagged, me and you would forgot something had already be done.

PR #218, opened 2022-06-10
Trainer profile: add validation on social network urls

maxime-poulain:
It is not longer possible to enter wrong url when a trainer edits its profile

https://smartjira.atlassian.net/browse/CFA-478


Victor-van-Duijnen (approved):
Nothing to add. Maybe next step we could check for specific social-account patterns (e.g. twitter => [http://twitter.com/[username]]) though it would probably be very very low priority XD

PR #219, opened 2022-06-13
Apply automatic column naming on enumerations

maxime-poulain:
Introduce a configuration on entities having one ore more enumeration that renames the db column name to the following:
* One enumeration property of a given type: [Enumeration Type] + "Id"
* Multiple enumeration properties of a give type : [Property Name] + "Id"


src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Persistence/EfCore/Extensions/EnumerationModelBuilderExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Smart.FA.Catalog.Infrastructure.Persistence.EfCore.Extensions;
+
+public static class EnumerationModelBuilderExtensions
+{
+    /// <summary>
+    /// Applies automatic conversion between enumeration values and their type.
+    /// </summary>
+    /// <param name="modelBuilder">The <see cref="ModelBuilder" /> on which is applied the automatic conversion.</param>
+    public static void ApplyConverterOnEnumerations(this ModelBuilder modelBuilder)
```
maxime-poulain, line 16: This was reviewed in a previous PR can be skipt.

LievenPraet (approved):

PR #239, opened 2022-06-22
Extract training creation/edition form into a partial view

maxime-poulain:
Simply move the HTML of the form to create or edit a training into a partial view.
This will help maintaining those two pages.

If somebody got a better naming for
[_TrainingCreationEditionForm.cshtml](https://github.com/smartcoop/smart-cfa/pull/239/files#diff-b43f5e5d2115aab83f76fa130af49942db3ac6369055dd9d1618a200557c3554) say so :).

This PR is required for https://smartjira.atlassian.net/browse/CFA-432

### DIFF
--- a/src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/Trainings/Update/UpdateTrainingViewModel.cs
+++ b/src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/Trainings/Update/UpdateTrainingViewModel.cs
@@ -3,35 +3,17 @@ using Smart.FA.Catalog.Application.UseCases.Queries;
 using Smart.FA.Catalog.Core.Domain.Dto;
 using Smart.FA.Catalog.Core.Domain.ValueObjects;
 using Smart.FA.Catalog.Shared.Domain.Enumerations.Training;
+using Smart.FA.Catalog.Web.Pages.Admin.Trainings.Create;
 
 namespace Smart.FA.Catalog.Web.Pages.Admin.Trainings.Update;
 
-public class UpdateTrainingViewModel
+public class UpdateTrainingViewModel : CreateTrainingViewModel
 {
-    public string? Title { get; init; }
-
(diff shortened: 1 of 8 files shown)


src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/Trainings/Update/UpdateTrainingViewModel.cs
```diff
@@ -3,35 +3,17 @@
 using Smart.FA.Catalog.Core.Domain.Dto;
 using Smart.FA.Catalog.Core.Domain.ValueObjects;
 using Smart.FA.Catalog.Shared.Domain.Enumerations.Training;
+using Smart.FA.Catalog.Web.Pages.Admin.Trainings.Create;
 
 namespace Smart.FA.Catalog.Web.Pages.Admin.Trainings.Update;
 
-public class UpdateTrainingViewModel
+public class UpdateTrainingViewModel : CreateTrainingViewModel
```
Victor-van-Duijnen, line 10: That's a bit strange to have the UpdateTrainingViewModel inherits from a CreateTrainingViewModel. If we go that route, we should have a common class TrainingViewModel and then a CreateTrainingViewModel and UpdateTrainingViewModel inheriting from that class.

Victor-van-Duijnen (approved):
I think the naming of the partial view is ok :)

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/Trainings/Update/UpdateTrainingViewModel.cs
maxime-poulain: In the first draft I did that, having a common class with the same fields.

Usually it is not a bad design per se to have the Edit inheriting from the Create but when you do that usually you have a
public int Id {get; set; } in the class itself which we don't have;

I can split it but then I ll do that in the next PR that is still in draft.

LievenPraet (approved):

PR #240, opened 2022-06-22
Add support for rich text editors.

maxime-poulain:
This PR is a follow up of https://github.com/smartcoop/smart-cfa/pull/239

This PR introduces a rich text editor for the creation/edition of trainings.
After discussion quill has been chosen.
A quick list / overview of existing popular library can be seen [here](https://github.com/jaredreich/pell#comparisons).

It would not be too much complicated to change to another library. 
However those point need to be discussed:

- What size should we give by default to those editors ?
- Do we need to customize a bit the css ?

Here is a quick demo how it looks like, for the time being.

https://user-images.githubusercontent.com/90202188/175110419-dba23ca3-2b38-47bf-9289-3a598b20c4b7.mp4


src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/Trainings/Create/Index.cshtml.cs
```diff
@@ -44,13 +40,13 @@ public async Task<IActionResult> OnPostAsync()
             return Page();
         }
 
-        var request = CreateTrainingViewModel.MapToRequest(UserIdentity.CurrentTrainer.Id, UserIdentity.CurrentTrainer.DefaultLanguage);
-        var reponse = await Mediator.Send(request);
+        var createTrainingRequest = CreateTrainingViewModel.MapToRequest(UserIdentity.CurrentTrainer.Id, UserIdentity.CurrentTrainer.DefaultLanguage);
+        var response = await Mediator.Send(createTrainingRequest);
```
LievenPraet, line 44: The var for the request has details in the name but the response does not.
We should choose 1 approach and be consistent with it.
```suggestion
        var request = CreateTrainingViewModel.MapToRequest(UserIdentity.CurrentTrainer.Id, UserIdentity.CurrentTrainer.DefaultLanguage);
        var response = await Mediator.Send(createTrainingRequest);
```

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/Trainings/Shared/_TrainingCreationEditionForm.cshtml
```diff
@@ -0,0 +1,192 @@
...
+            });
+
+        editor.on('text-change',
+            function() {
+                var text = editor.root.innerHTML;
+                var hidden = document.getElementById(hiddenValueElement);
+                hidden.value = text;
+            });
+    }
+</script>
```
LievenPraet, line 186: We need to eventually create a RTE component in the design library and move the js there

LievenPraet (changes requested):

Victor-van-Duijnen:
You can fix the broken bold font of the text editor by adding the css
 ` .ql-editor strong{
         font-weight:bold;
     }`

src/UserAdmin/src/Smart.FA.Catalog.Web/wwwroot/js/lib.js
```diff
@@ -63,11 +63,16 @@ function RegisterCultureChange() {
...
-        forms[i].addEventListener('keydown', ignoreEnterKeyEventHandler, true);
-        forms[i].addEventListener('keyup', ignoreEnterKeyEventHandler, true);
+        forms[i].addEventListener('keydown', disableEnterKeyEvent, true);
+        forms[i].addEventListener('keyup', disableEnterKeyEvent, true);
     }
 
     function disableEnterKeyEvent(e) {
+        var e = (e) ? e : ((event) ? event : null);
+        var node = (e.target) ? e.target : ((e.srcElement) ? e.srcElement : null);
+        if (node.nodeName == "div" || node.nodeName == "DIV") {
```
Victor-van-Duijnen, line 73:
```suggestion
        if (node.nodeName === "div" || node.nodeName === "DIV") {
```

Victor-van-Duijnen: Do we want to be able to send a form with the enter key from a div like the title of a training?

src/UserAdmin/src/Smart.FA.Catalog.Web/wwwroot/js/lib.js
```diff
@@ -63,11 +63,16 @@ function RegisterCultureChange() {
...
     for (var i = 0; i < forms.length; i++) {
-        forms[i].addEventListener('keydown', ignoreEnterKeyEventHandler, true);
-        forms[i].addEventListener('keyup', ignoreEnterKeyEventHandler, true);
+        forms[i].addEventListener('keydown', disableEnterKeyEvent, true);
+        forms[i].addEventListener('keyup', disableEnterKeyEvent, true);
     }
 
     function disableEnterKeyEvent(e) {
+        var e = (e) ? e : ((event) ? event : null);
+        var node = (e.target) ? e.target : ((e.srcElement) ? e.srcElement : null);
```
Victor-van-Duijnen, line 72: Suggested by the IDE: 
```suggestion
     var e = (e) ? e : ((Event) ? Event : null);
        var node = (e.target) ? e.target : ((e.target) ? e.target : null);
```
But more seriously, I seem to have the same behaviour with or without the method (tested on the training page only in Chrome and FIrefox in private mode). Is the enter key event blocked by default with the rich text editor fields (in quill) . Just mentionning it because I found it strange, is it the same for you @maxime-poulain ?

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/Trainings/Shared/_TrainingCreationEditionForm.cshtml
```diff
@@ -0,0 +1,192 @@
...
+@using Smart.Design.Razor.TagHelpers.Spacer
+@using Smart.FA.Catalog.Core.Services
+@using Smart.FA.Catalog.Shared.Domain.Enumerations.Common
+@using Smart.FA.Catalog.Shared.Domain.Enumerations.Training
+@model Smart.FA.Catalog.Web.Pages.Admin.Trainings.Create.CreateTrainingViewModel
+@inject IStringLocalizer<CatalogResources> Localizer
+@{
+    var identityService = ViewContext.HttpContext.RequestServices.GetRequiredService<IUserIdentity>();
+
+    var superUser = identityService.IsSuperUser;
```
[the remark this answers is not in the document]
maxime-poulain: Was hard to think properly yday x).
I am gonna inject directly the service instead of using the service locator

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/Trainings/Shared/_TrainingCreationEditionForm.cshtml
maxime-poulain: Yes but I guess let's wait to see if emily has something else to propose ?

src/UserAdmin/src/Smart.FA.Catalog.Web/wwwroot/js/lib.js
```diff
@@ -41,7 +41,7 @@ function RegisterScrollToFirstElementInError() {
 // Handling of the culture switching.
 document.addEventListener("DOMContentLoaded",
     function() {
-        RegisterCultureChange();
+        //RegisterCultureChange();
```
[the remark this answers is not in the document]
maxime-poulain: Yup it looks for a <select> that doesnt exist (at least it is commented too).

src/UserAdmin/src/Smart.FA.Catalog.Web/wwwroot/js/lib.js
maxime-poulain: It's just to have the break line working.

src/UserAdmin/src/Smart.FA.Catalog.Web/wwwroot/js/lib.js
maxime-poulain: Hm we loose the falback to srcElement doing so

chastelain:
- Isn't the question of character limitation a database-side problematic?
- Allowing rich-text input, is XSS being taken into account? In DeskAPI we use this package to clean 
text inputs: https://github.com/Vereyon/HtmlRuleSanitizer
- > but its flaws:
"you must arrrive >>>>>>>>>>>>>>>>>> there";
the >>>>>>>> would be removed.

You sure?

image

maxime-poulain:
> Isn't the question of character limitation a database-side problematic?
What do you exactly mean ? With the correct validation it would not 

>  Allowing rich-text input, is XSS being taken into account?
No, not at all ! I will look into that 

> You sure?
Heh, yes, I didn't post the actual example I tried x)
image

chastelain:
> > Isn't the question of character limitation a database-side problematic?

> What do you exactly mean ? With the correct validation it would not

I mean historically in our databases, text columns have always been constrained to a max fixed value, I don't know if it's something that has to be considered for CFA or not, but with no limitation, no problem.
However, if you strip the HTML on the input, aren't you saving the text in the DB without the formatting? How would you "unstrip" it back?
Or is it just to bypass an input validation on char counting?

> You sure?

>> Heh, yes, I didn't post the actual example I tried x)

Didn't try with `<` inside html text tags xD My bad.

Victor-van-Duijnen:
@chastelain It's only for validation, the actual string stored will contain the metadata but it's true that, besides the actual content of a RTF, we should maybe also validate the size of the text + metadata (someone could send really really big strings full of metadata but no actual text). For example, we need to have a minimum of 30 characters for some fields (business rule). If we make the check on the complete string, it won't work properly.

@maxime-poulain Looking at quill's issues, I found that thread: https://github.com/quilljs/quill/issues/1918 and they seem to recommend https://github.com/apostrophecms/sanitize-html to prevent XSS.

It's true that it doesn't look that easy to find a regex that only take the string between matching html tags :'(

maxime-poulain:
> @chastelain It's only for validation, the actual string stored will contain the metadata but it's true that, besides the actual content of a RTF, we should maybe also validate the size of the text + metadata (someone could send really really big strings full of metadata but no actual text). For example, we need to have a minimum of 30 characters for some fields (business rule). If we make the check on the complete string, it won't work properly.
> 
> @maxime-poulain Looking at quill's issues, I found that thread: [quilljs/quill#1918](https://github.com/quilljs/quill/issues/1918) and they seem to recommend https://github.com/apostrophecms/sanitize-html to prevent XSS.
> 
> It's true that it doesn't look that easy to find a regex that only take the string between matching html tags :'(

We need to have the sanitization at the back end level, quil js already escapes, however, the things that could lead into a XSS.

maxime-poulain:
@Victor-van-Duijnen 
@chastelain 
@LievenPraet 
Sup guys,

after struggling, I finally seem to reach an acceptable solution.

# First, summary of issues:
The deskAPI uses a custom resolver that applies sanitization on every string but one marked with a given attribute (something like [IgnoreSanitization].

Sadly, and as expected x), we cannot use that approach (at least with System.Text.Json, I think it could have been done with Newton but the best is not to use that lib anymore).

* We are posting forms and the actual type are form data and not JSON.
* Starting to manually apply sanitization on properties seems to be retarded.

# Solution:

I don't think having everything sanitized is a good way we might break stuff that are not intended to.
After some digs, the best library to do sanitization is https://github.com/mganss/HtmlSanitizer

I could figure out how to add a custom ModelBinder and Provider that changes the default behavior of binding on string that have an attribute

> [Sanitized]

So the API that PR might introduce looks like this:

```csharp
public class EditProfileCommand : IRequest<ProfileEditionResponse>
{
    public int TrainerId { get; set; }

    [Sanitized]
    public string? Bio { get; set; }

    [Sanitized]
    public string? Title { get; set; }

    public Dictionary<int, string>? Socials { get; set; }
}
```

Both **Bio** and **Title** get sanitized


# Last: validation, what to do with that ! ?
Now we are posting HTML  + CSS attribute the strings become super long but the validation still needs to be working properly.

There's two ways I could identity:

1.  Simple Regex escaping ( <div>hello toto <<<<<<<<<<< ?? </div> will result in hello toto, heh.
2. Track HTMLAgilityPack that will do that properly. (We get like a factory for a simple thing tho ;( );


 ## Comments ?

Victor-van-Duijnen:
@maxime-poulain For sanization I am not sure what were the criteria for chosing that specific library but I will trust your choice on it. I would indeed not have recommended sanization everywhere by default and custom model binding seems right to me.

For the validation part, HtmlAgilityPack might be the best solution when it comes to result but it also comes with a lot of baggages which I am not too fond of. Just to be sure I understand we can not do much better than the regex that will scrap everything after "<" even if it's not a tag? If we really can't do otherwise, then HtmlAgilityPack...

maxime-poulain:
> @maxime-poulain For sanization I am not sure what were the criteria for chosing that specific library but I will trust your choice on it. I would indeed not have recommended sanization everywhere by default and custom model binding seems right to me.
> 
> For the validation part, HtmlAgilityPack might be the best solution when it comes to result but it also comes with a lot of baggages which I am not too fond of. Just to be sure I understand we can not do much better than the regex that will scrap everything after "<" even if it's not a tag? If we really can't do otherwise, then HtmlAgilityPack...

Regarding the library there aren't out there. However it is the most recent one and it is very flexible.
There was the one they used in the DeskApi but it has not been updated for more than one year.

Back to the current lib:
image

I went with the default config maybe in the future that might need to be adjusted 

Regarding validation:

> hello toto <<<<<<<<<<< ??

will become 

> hello toto

I don't mind going with the regex it is faster. Note that one of the reason i did not want to go with DeskApi's one it is because it uses HtmlAgilityPack under the hood x).

I can write some kind of benchmark also to compare regex vs HTmlPack to validate the number of chars.

chastelain:
IMO you're focusing on performance when there's no real need and it's no game changer.

Using regex will be faster but in return the output may not be correct. 
It just takes one bug ticket from a user who really wants to format its text with `<<<<<` to make us change its behaviour lately xD

Is it really necessary to sanitize? ASP Core doesn't offer already some mechanisms? https://docs.microsoft.com/en-us/aspnet/core/security/cross-site-scripting?view=aspnetcore-6.0

Is it really bad to use HAP?

maxime-poulain:
We need to use some ways to decode HTML (HtmlString or @HTml.Raw) to render correctly the HTML and not the escaped one.

QuilLJS escapes already for use <scirpt> to &:ltscript&:gt but this is not enough as anyone could simply use postman or something else to send some data.

like stated : ASP.NET Core MVC provides an HtmlString class which isn't automatically encoded upon output. This should never be used in combination with untrusted input as this will expose an XSS vulnerability.

https://user-images.githubusercontent.com/90202188/176441359-168d5aed-4cfd-4737-94dc-e4c3114b508f.mp4

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Admin/Trainings/Create/Index.cshtml.cs
maxime-poulain: That was updated and set a meaningful name.

src/UserAdmin/src/Smart.FA.Catalog.Web/wwwroot/js/lib.js
maxime-poulain: I see i have adjusted a bit the JS code based on this feedbck.

chastelain (approved):

LievenPraet (approved):

chastelain:
>> IMO you're focusing on performance when there's no real need and it's no game changer.
>> 
>> Using regex will be faster but in return the output may not be correct. It just takes one bug ticket from a user who really wants to format its text with <<<<< to make us change its behaviour lately xD
>> 
>> Is it really necessary to sanitize? ASP Core doesn't offer already some mechanisms? https://docs.microsoft.com/en-us/aspnet/core/security/cross-site-scripting?view=aspnetcore-6.0
>> 
>> Is it really bad to use HAP?

>> 

> @chastelain For the record, I was not suggesting that we use a sub-par (namely the regex mentioned) solution because I was afraid of the performance of HtmlAgilityPack, I wanted to confirm that there wasn't other regex/way to cover that use case, without having to import a library that contains a lot of things we don't need.

Mmmm ok, but my comment was addressed to the analysis Maxime did in https://github.com/smartcoop/smart-cfa/pull/240#issuecomment-1168334755 😅

src/Shared/Smart.FA.Catalog.Shared/Helper/Guard.cs
```diff
@@ -43,6 +45,19 @@ public static string AgainstInvalidEmail(string? email, string parameterName, st
         return input;
     }
 
+    public static string? AgainstMaxHtmlInnerTextLength(string? htmlInput, int maxValue, [CallerArgumentExpression("htmlInput")] string paramName = null, string? message = null)
```
Victor-van-Duijnen, line 48: Not fan of putting the HTMLAgilityPack inside the shared application because now showcase has a depency to a package that is completely useless for it. Was it not possible to pass the inner html text to the AgainstMaxLength instead of having a custom guard clause for it?

src/UserAdmin/src/Smart.FA.Catalog.Application/SeedWork/Attributes/SanitizedAttribute.cs
```diff
@@ -0,0 +1,9 @@
+namespace Smart.FA.Catalog.Application.SeedWork.Attributes;
+
+/// <summary>
+/// Specifies that a property should be sanitized.
```
Victor-van-Duijnen, line 4: "Should be" or "will be" sanitized?

src/UserAdmin/src/Smart.FA.Catalog.Application/SeedWork/Attributes/SanitizedAttribute.cs
```diff
@@ -0,0 +1,9 @@
...
+/// </summary>
+[AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
```
Victor-van-Duijnen, line 6: The AllowMultiple option is the default apparently, so it can be removed.

src/UserAdmin/src/Smart.FA.Catalog.Application/SeedWork/Attributes/SanitizedAttribute.cs
maxime-poulain: Without a "able" suffix it does not make any sense to have "should".
Btw [Sanitizable] was my first go.

If [Sanitize] is kept.

> Specifies that a property is to be sanitized

is that correct in english ? 
@LievenPraet  inputs ?

maxime-poulain:
@Victor-van-Duijnen 
@LievenPraet 

We will most likely not merge this PR because we are facing too many issues with Design System's CSS that interfere will basic HTML.

Example 

> <ul> <li> </li> </ul> 

are not rendered and many more visual issue.
There's workaround tho like manipulating the output to add accordingly correct classes but I guess it starts to be way to many workarounds.

I suggest to have a chat with Joelle tho to see what's her opinion.
The backend part can be merged or the PR can just be closed and later on you can see what to do about it.

PR #241, opened 2022-06-28
SuperUser page for denying access to a trainer 

Victor-van-Duijnen:
[This PR](https://smartjira.atlassian.net/browse/CFA-468) introduces a super user page and a set of method to list trainers, delete their data in the database and/or deny/allow them access. 

As always the wording used are up for discussion (I know for example inclusive wording would replace blacklist by denylist or rejectlist or blocklist but I chose to take the traditional route to avoid confusion).


maxime-poulain:
I feel like most (or all?) VC are overkill for their uses and could be replaced with partial view.
The VC we already have (AdminTrainingTile) make sense because it has a bit of logic (even tho it is more or less not used).

src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/GetOtherTrainersListQuery.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    public async Task<PagedList<Trainer>> Handle(GetOtherTrainersListRequest request, CancellationToken cancellationToken)
+    {
+        var trainerQueryable = _catalogContext.Trainers
+            .Include(trainer => trainer.SocialNetworks)
+            .Include(trainer => trainer.Approvals)
+            .Include(trainer => trainer.Assignments)
+            .Where(trainer => trainer.Id != request.SelfTrainerId);
+
+        if (!request.TrainerName.IsNullOrEmpty())
```
maxime-poulain, line 29: You should not be using this, it's meant to be used on collections:
image

src/Shared/Smart.FA.Catalog.Shared/Collections/PagedList.cs
```diff
@@ -79,3 +79,8 @@ public PageItem(int currentPage, int pageSize)
         PageSize = pageSize;
     }
 }
+
+public static class PagedListExtensions
```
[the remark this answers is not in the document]
Victor-van-Duijnen: Yes actually that class even existed already moved it just before your comment XD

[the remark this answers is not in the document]
maxime-poulain: But I feel it is not necessary to have that (as far I understood it is used for the tile pagination view ?)

We should merge the pagination component once and for all. You can even try it out quickly in like 15 minutes it's done (I can also give you my linqpad script to populate hundreds of trainings).

You can also introduce some interface that exposes the properties you wanted access to

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/SuperUser/Trainers/SuperUserTrainerList.cshtml
```diff
@@ -0,0 +1,131 @@
...
+            <div class="o-card-grid small-up-1">
+
+                @foreach (var trainer in Model.TrainerList)
+                {
+                    @await Component.InvokeAsync(nameof(Shared.Components.SuperUserTrainerListTile), new { trainer })
+                }
+            </div>
+
+            // Paging interface
+            @await Component.InvokeAsync(nameof(Shared.Components.PagingTile.PagingTile), new { list = Model.TrainerList.ToPagedListOfObjects(), route = Routes.SuperUserTrainerList })
```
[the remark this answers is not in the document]
Victor-van-Duijnen: I don't really see the point to have a view component named differently from the class of the view component. Following the doc, they are advising against using hardcoded string https://docs.microsoft.com/en-us/aspnet/core/mvc/views/view-components?view=aspnetcore-6.0#avoid-hard-coded-strings

src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/GetOtherTrainersListQuery.cs
Victor-van-Duijnen: Yes good catch, went too fast on that one.

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/SuperUser/Trainers/SuperUserTrainerList.cshtml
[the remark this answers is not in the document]
maxime-poulain: In general it is a bad idea to hard code string instead of using constants.
you could for example have a ViewComponents class that holds all the nameof(xxxx) where xxx are the VC classes.
For example, none of the tag helper uses hardcoded string but constants (well they are hardcoded).

Regarding the VC:

image
image

That comment about avoiding hardcoded string is because you may rename your class name and then you got fucked.

image

What's the behavior with rider ?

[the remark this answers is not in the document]
maxime-poulain: That's the trade off between the two choices. The compilation doesn't break so if you rename you might have some kind of surprises

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Shared/Components/SuperUserTrainerListTile/Default.cshtml
```diff
@@ -0,0 +1,33 @@
...
+            {
+                <smart-muted-text text="@($"{socialNetwork.SocialNetwork.Name}: {socialNetwork.UrlToProfile}")"></smart-muted-text>
+            }
+        }
+    </smart-card-body>
+    <smart-card-actions>
+        <div class="o-flex o-flex--vertical-center o-flex--justify-between o-flex--spaced-wide">
+            <div class="c-button-toolbar">
+                <form id="delete-trainer-form-@Model.Id" method="post" asp-page-handler="Delete" asp-route-id="@Model!.Id">
+                    <button class="c-button c-button--borderless c-button--icon" type="button" data-trainerId="@Model.Id" data-dialog="delete-trainer-dialog-confirmation" onclick="setTrainerId(@Model.Id)">
```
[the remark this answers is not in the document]
LievenPraet: We need a confirm smart button that could be used here in the future.
You agree, @maxime-poulain ?

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Shared/Components/SuperUserTrainerListTile/SuperUserTrainerListTile.cs
```diff
@@ -1,12 +1,43 @@
...
+
+public class TrainerListTile
+{
+    public int Id { get; set; }
+
+    public string Email { get; set; } = null!;
+
+    public Name Name { get; set; } = null!;
+
+    public Dictionary<string, string> SocialNetwork { get; set; } = new();
```
maxime-poulain, line 38: You could pluralize this property.

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/SuperUser/Trainers/SuperUserTrainerList.cshtml.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                // Delete data related to the trainer
+                await Mediator.Send(new DeleteTrainerRequest { TrainerId = id });
+                // Display success message
+                TempData.AddGlobalAlertMessage(CatalogResources.TrainerDeletedWithSuccess, AlertStyle.Success);
+                return RedirectToPage();
+            }
+        }
+        TempData.AddGlobalAlertMessage(CatalogResources.UnExpectedError, AlertStyle.Error);
+        return RedirectToPage();
```
LievenPraet, line 68: It seems more logical to handle the error case first and put the success case last.

LievenPraet (changes requested):

src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Commands/DeleteTrainerCommand.cs
```diff
@@ -0,0 +1,35 @@
...
+    public DeleteTrainerCommand(CatalogContext catalogContext)
+    {
+        _catalogContext = catalogContext;
+    }
+
+    public async Task<DeleteTrainerResponse> Handle(DeleteTrainerRequest request, CancellationToken cancellationToken)
+    {
+        DeleteTrainerResponse response = new();
+        var trainer = await _catalogContext.Trainers.FirstOrDefaultAsync(trainer => trainer.Id == request.TrainerId, cancellationToken);
+        _catalogContext.Remove(trainer);
```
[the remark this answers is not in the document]
maxime-poulain: A golden rule (to me at least) should be that

> any commands (i.e. something that will change the state of your system) should(must?) come with a validator associated with it.

That way it makes it easier to build a more resilient system.

I think it is pretty straight forward (at least for non complex scenerios): 

- "I need to delete some trainer";
- "Ok writing the validation what do I need to be true in order to process ? "
- "The trainer must exist"

Then you add a rule that ensures the id you provided does exist.

Of course nothing prevents you to add on top of that some exceptions or debug assertions in your BL / handlers.

Note that if the validation logic is present the framework will handle for you the end user feedback.

Second note, it is easy to consider having an exception being thrown if there are no validator associated to a command.
(Well one could add a validator that just do nothing but it could be a guard to be considered).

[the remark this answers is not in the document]
maxime-poulain: That's why it becomes interesting with using EF, instead of making a call (Any()) to see if an entity does exist you can just issue the retrieval and it will be cached.
So later on the same object will be resolved instead of hitting the database again.

Performance wise even if you would be doing some select 1 to make sure something exist it would not matter that much in the end.

src/UserAdmin/src/Smart.FA.Catalog.Web/Authorization/Handlers/UserAdminAuthorizationResultHandler.cs
```diff
@@ -20,10 +20,14 @@ public class UserAdminAuthorizationResultHandler : IAuthorizationMiddlewareResul
...
     {
         _userIdentity = context.RequestServices.GetRequiredService<IUserIdentity>();
+        var authorizationFailure = authorizeResult.AuthorizationFailure!;
 
         if (!_userIdentity.IsSuperUser && authorizeResult.Forbidden)
         {
-            var authorizationFailure = authorizeResult.AuthorizationFailure!;
+            if (authorizationFailure.FailedRequirements.AnyOfType<MustNotBeBlackListedRequirement>())
+            {
+                context.Response.StatusCode = StatusCodes.Status403Forbidden;
```
maxime-poulain, line 29: If it possible the StatusCode is already that value ?

src/UserAdmin/src/Smart.FA.Catalog.Web/Authorization/Policy/Requirements/MustNotBeBlackListed.cs
```diff
@@ -0,0 +1,33 @@
+using MediatR;
+using Microsoft.AspNetCore.Authorization;
+using Smart.FA.Catalog.Application.UseCases.Queries;
+using Smart.FA.Catalog.Core.Services;
+
+namespace Smart.FA.Catalog.Web.Authorization.Policy.Requirements;
+
+/// <summary>
+/// Requires that the user was not blacklisted.
```
maxime-poulain, line 9:
```suggestion
/// Requires that the user is not blacklisted.
```

src/UserAdmin/src/Smart.FA.Catalog.Web/Extensions/MvcBuilderExtensions.cs
```diff
@@ -21,7 +18,10 @@ private static void AddPageConventionsAuthorization(this PageConventionCollectio
         conventions.AuthorizePage("/Admin/Trainings/Update/Index", Policies.MustBeSuperUserOrTrainingCreator);
 
         // Folders
+        conventions.AuthorizeFolder("/Admin", Policies.MustNotBeBlackListed);
+        conventions.AuthorizeFolder("/SuperUser", Policies.MustNotBeBlackListed);
```
maxime-poulain, line 22: Should we be able to black list a super user ?
I guess by default a user that can be blacklisted should be able to access the super user space ?

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Shared/Components/SuperUserTrainerListTile/SuperUserTrainerListTile.cs
```diff
@@ -0,0 +1,51 @@
...
+    private readonly IMediator _mediator;
+
+    public SuperUserTrainerListTile(IMediator mediator)
+    {
+        _mediator = mediator;
+    }
+
+    public async Task<IViewComponentResult> InvokeAsync(Trainer trainer)
+    {
+        var isTrainerBlackListed = await IsBlackListed(trainer);
```
maxime-poulain, line 21:
```suggestion
        var isTrainerBlackListed = await IsBlackListedAsync(trainer);
```

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/Shared/Components/SuperUserTrainerListTile/SuperUserTrainerListTile.cs
```diff
@@ -0,0 +1,51 @@
...
+            Id = trainer.Id,
+            Name = trainer.Name,
+            ApplicationType = ApplicationType.FromValue(trainer.Identity.ApplicationTypeId).Name,
+            UserId = trainer.Identity.UserId,
+            IsBlackListed = isTrainerBlackListed
+        };
+        return View(trainerListTile);
+    }
+
+    public async Task<bool> IsBlackListed(Trainer trainer) => await _mediator.Send(new IsTrainerBlackListedRequest { TrainerId = trainer.Id });
```
maxime-poulain, line 35:
```suggestion
    public async Task<bool> IsBlackListedAsync(Trainer trainer) => await _mediator.Send(new IsTrainerBlackListedRequest { TrainerId = trainer.Id });
```

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/SuperUser/Trainers/SuperUserTrainerList.cshtml.cs
```diff
@@ -0,0 +1,118 @@
...
+{
+    [BindProperty(SupportsGet = true)]
+    public string? TrainerNameOrEmailFilterQuery { get; set; }
+
+    [BindProperty(SupportsGet = true)]
+    public int CurrentPage { get; set; } = 1;
+
+    public IUserIdentity UserIdentity { get; }
+
+    public IMediator Mediator { get; }
```
maxime-poulain, line 26: Do you need those two properties to be public ?

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/SuperUser/Trainers/SuperUserTrainerList.cshtml.cs
```diff
@@ -0,0 +1,118 @@
...
+        Mediator = mediator;
+        Settings = superUserOptions.Value;
+    }
+
+    public async Task OnGetAsync()
+    {
+        // Display side panel as selected
+        ViewData[nameof(SuperUserSideMenuItem)] = SuperUserSideMenuItem.SuperUserTrainerList;
+
+        // Get All trainers (except yourself) paged
```
maxime-poulain, line 44: Yourself would refer to the reader of the code I guess :p

```suggestion
        // Get All trainers (except current connected user) paged
```

or

```suggestion
        // Get All trainers (but current connected user) paged
```

or keep it that way.

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/SuperUser/Trainers/SuperUserTrainerList.cshtml.cs
```diff
@@ -0,0 +1,118 @@
...
+        var getTrainerListRequest =
+            new GetOtherTrainersListRequest { TrainerNameOrEmailQueryFilter = TrainerNameOrEmailFilterQuery, PageItem = new PageItem(CurrentPage, Settings.NumberOfTrainersPerPage), SelfTrainerId = UserIdentity.Id };
+        TrainerList = await Mediator.Send(getTrainerListRequest);
```
maxime-poulain, line 47: Having the code a bit formatted would be better.

maxime-poulain (changes requested):

src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/GetTrainerQuery.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        GetTrainerResponse response = new();
+        response.Trainer = await _catalogContext.Trainers.SingleAsync(trainer => trainer.Id == request.TrainerId, cancellationToken);
+        response.SetSuccess();
+
+        return response;
+    }
+}
+
+public class GetTrainerRequest : IRequest<GetTrainerResponse>
```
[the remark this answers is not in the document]
LievenPraet: `ById` is preferable

LievenPraet (changes requested):

src/UserAdmin/src/Smart.FA.Catalog.Infrastructure/Migrations/ContextModelSnapshot.cs
```diff
@@ -18,11 +18,26 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                .HasAnnotation("ProductVersion", "6.0.6")
                 .HasAnnotation("Relational:MaxIdentifierLength", 128);
 
             SqlServerModelBuilderExtensions.UseIdentityColumns(modelBuilder, 1L, 1);
 
+            modelBuilder.Entity("Smart.FA.Catalog.Core.Domain.Authorization.BlackListedTrainer", b =>
+                {
+                    b.Property<int>("TrainerId")
+                        .ValueGeneratedOnAdd()
+                        .HasMaxLength(200)
```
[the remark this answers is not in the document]
Victor-van-Duijnen: Yes indeed, some weird swapping happened. I have corrected that.

src/UserAdmin/src/Smart.FA.Catalog.Web/Authorization/Handlers/UserAdminAuthorizationResultHandler.cs
Victor-van-Duijnen: Yes but in that case, I would rather be explicit about what to expect in that kind of failure.

src/UserAdmin/src/Smart.FA.Catalog.Web/Extensions/MvcBuilderExtensions.cs
Victor-van-Duijnen: I have asked business, I will come back to that one.

maxime-poulain (changes requested):

src/UserAdmin/src/Smart.FA.Catalog.Web/Pages/403.cshtml.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        else if (HttpContext.Items.ContainsKey(ErrorTitleKey))
+        {
+            ErrorTitle = HttpContext.Items[ErrorTitleKey]!.ToString()!;
+        }
+
+        else
+        {
+            ErrorTitle = CatalogResources.Sorry;
+        }
```
LievenPraet, line 38: Same remark here. Code would be cleaner with returns instead of if/else...

src/UserAdmin/src/Smart.FA.Catalog.Web/Extensions/MvcBuilderExtensions.cs
Victor-van-Duijnen: I have confirmation from @Geraldineb1  that a super user could be blacklisted.

LievenPraet (approved):

src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/GetOtherTrainersListQuery.cs
```diff
@@ -0,0 +1,46 @@
...
+                .Where(trainer => trainer.Name.FirstName.Contains(request.TrainerNameOrEmailQueryFilter!)
+                    || trainer.Name.LastName.Contains(request.TrainerNameOrEmailQueryFilter!)
+                    || trainer.Email!.Contains(request.TrainerNameOrEmailQueryFilter));
+        }
+
+        return await trainerQueryable.PaginateAsync(request.PageItem, cancellationToken);
+    }
+}
+
+public class GetOtherTrainersListRequest : IRequest<PagedList<Trainer>>
```
maxime-poulain, line 41: Not requesting a change just a suggestion, so as you prefer:

> GetTrainersByNameOrEmailExceptSelf ?

src/UserAdmin/src/Smart.FA.Catalog.Application/UseCases/Queries/GetOtherTrainersListQuery.cs
```diff
@@ -0,0 +1,46 @@
...
+    private readonly CatalogContext _catalogContext;
+
+    public GetOtherTrainersListQuery(CatalogContext catalogContext)
+    {
+        _catalogContext = catalogContext;
+    }
+
+    public async Task<PagedList<Trainer>> Handle(GetOtherTrainersListRequest request, CancellationToken cancellationToken)
+    {
+        var trainerQueryable = _catalogContext.Trainers
```
maxime-poulain, line 23: Since we have no validator would be wise to check if the self trainer id is not equal to default ?

maxime-poulain (approved):
Left one optional minor comment

PR #247, opened 2022-07-05
Remove "handler" postfix in naming

Victor-van-Duijnen:
/!\ first require merging of PR #245

Since we are on the topic of quick refactoring, I also wanted to remove the postfix "Handler" on our commands/queries to have them all follow the same naming guideline. I guess it might be a good time to review their names, too, but let's keep this PR to that scope (so only queries/command renaming), please.


LievenPraet:
File name `GetTrainersFromTrainingQuery.cs` should be `GetTrainersByTrainingQuery.cs`. Same remark for other file names.

LievenPraet (changes requested):
File name GetTrainersFromTrainingQuery.cs should be GetTrainersByTrainingQuery.cs. Same remark for other file names.

maxime-poulain:
I think the query names should be explicit as possible on which fields it gonna look for.

Example

> GetTrainingsByTitle

> GetTrainingByCreatorId

> etc

However, finding the correct name is not always a easy task .

In the case of 

> GetTrainersFromTrainingQuery 

or 

> GetTrainersByTrainingQuery

I guess none of those is relevant as it actually gets every trainings that a trainer is assigned to.

Victor-van-Duijnen:
> I think the query names should be explicit as possible on which fields it gonna look for.
> 
> Example
> 
> > GetTrainingsByTitle
> 
> > GetTrainingByCreatorId
> 
> > etc
> 
> However, finding the correct name is not always a easy task .
> 
> In the case of
> 
> > GetTrainersFromTrainingQuery
> 
> or
> 
> > GetTrainersByTrainingQuery
> 
> I guess none of those is relevant as it actually gets every trainings that a trainer is assigned to.

As it is not even used I will remove that handler alltogether.

maxime-poulain:
I would add that Command suffixes (and queries) should be the actual suffixes names suffixes of requests.
Classes that are handlers should be suffixed with handlers.
 ```csharp
public class PurchaseItemCommand : Command / ICommand / Whatever
{
    public int ItemId {get; set; }
}

public class PurchaseItemCommandHandler : CommandHandler / etc
{
   // do some stuff ..
}
```
There's some interesting paper by Greg Young.
He is a great speaker and architect, I advise checking [his talk](https://www.youtube.com/results?search_query=greg+young)s btw


https://cqrs.files.wordpress.com/2010/11/cqrs_documents.pdf

relevant point is at the bottom of page 11

Victor-van-Duijnen:
I don't disagree with using directly "_Command_" for our request but even Greg Young is following his example with the side note:
"_(...) the example in Listing 1 includes the pattern name after the name of the Command. This is
a decision that has many positives and negatives both linguistically and operationally. The choice over
whether to use a pattern name in a class name is one that should not be taken lightly by a development
team._"
So if we do that (particularly with the mediator implementation of Jimmy Bogaert who uses request/response terminilogy for its interfaces), it must a concerted decision.

maxime-poulain:
Well he meant by that

> CreateOrder vs CreateOrderCommand

I had already that discussion a while back with @BomUbik about having or not some suffix. But to me, having a suffix shows clearly the intend and help to avoid confusion.

CQS pattern, which is at the core of this application, should prime over some simple library API that's just a tool to achieve the goal (i.e. CQS);

Speaking of MediatR, have a look at this:

https://github.com/jbogard/MediatR/commit/cd2fc954a7d3fa860a551af9e345302b3714fac2

The reason the author made it was to make it easy to do VSA & CQS/CQSRS/VSA nothing more.
At some point he changed the API so it could be ore generic and "useable" in other use cases.

Victor-van-Duijnen:
I meant to say that we are stuck with IRequest and IRequestHandler as the name of the interfaces  which are implemented in our commands/query and handlers, since they are part of the IMediatr nuget. Those terms are not going anywhere (even if you create a custom interface implementing IRequestHandler/IRequest, it's just pushing the term one level deeper), even if they are incorrect. So if you have a xxxCommand (name of a pattern) implemeting IRequest (request is another pattern), now you have introduced a new term (command) that you have to deal with.
 I am not saying it's wrong, I am just saying it has a mental cost.

maxime-poulain:
What is exactly the subject of the topic, removing Handler from handlers of the commands / queries / requests ?

---

To answer to the above statement:

> So if you have a xxxCommand (name of a pattern) implementing IRequest (request is another pattern), now you have introduced a new term (command) that you have to deal with.

Yes you are right, that's why having specific interfaces for commands and queries is the way to go.
We should not care about the fact MediatR's API uses IRequest, it makes no sense to do so when the underlying architecture is based on CQS.

This could be debatable if this project was an API and we would do REPR, that's not the case.
This is just about the naming but even technically it brings more advantages.

Regardless of outcome, it is time, indeed, to harmonize the name of classes.

Victor-van-Duijnen:
While the harmonization of name in the handlers is definitively in the pipe, it's outside of the scope of this PR so I will be merging that now and opening another one for handlers after discussing it in a COP.

PR #256, opened 2022-07-12
[ShowCase] Fix reload page after send email to trainer on training details page

NourJammal:
Fix reload page after send email to trainer on training details page


Victor-van-Duijnen (approved):
Haha yes that makes more sense like that

PR #258, opened 2022-07-13
Add InquirySendEmailService with mail redirect

Victor-van-Duijnen:
Use a specific test mail (product@smart.coop) address for testing mail features for testing in development and staging.
Relating to ticket [cfa-500](https://smartjira.atlassian.net/browse/CFA-500)


src/Showcase/src/Smart.FA.Catalog.Showcase.Web/appsettings.Development.json
```diff
@@ -9,5 +9,9 @@
   "ConnectionStrings": {
     "Catalog": "Server=(LocalDB)\\MSSQLLocalDB; Database=Catalog; Integrated Security=true;",
     "Hangfire": "Server=(LocalDB)\\MSSQLLocalDB; Database=Catalog; Integrated Security=true;"
+  },
+  "Inquiry": {
+    "DefaultEmail": "product@smart.coop", // Email to whom will be sent the inquiries.
```
LievenPraet, line 14: The email address `product@smart.coop` has no indication that it is used for testing.
Would it be better to use an address that clearly shows that it is a "test address"?

src/Showcase/src/Smart.FA.Catalog.Showcase.Infrastructure/Mailing/Inquiry/Trainer/TrainerInquirySendTestRecipientEmailService.cs
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Smart.FA.Catalog.Showcase.Domain.Common.Options;
+
+namespace Smart.FA.Catalog.Showcase.Infrastructure.Mailing.Inquiry.Trainer;
+
+/// <summary>
+/// Test Trainer Inquiry mail service with sending of the mail to a specific address used for debugging and general testing.
+/// </summary>
+public class TrainerInquirySendTestRecipientEmailService : TrainerInquirySendEmailService
```
LievenPraet, line 12: The name `TrainerInquirySendTestRecipientEmailService` is a bit long and might get confusing.
Could you think about an alternative name that is less long?

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/appsettings.Development.json
maxime-poulain: Actually the code forces the recipient whatever the underlying trainer is.
If the trainer is Lieven Praet the mail will be sent to product@smart.coop.
I think having "test" or "redirect" in names are confusing and don't reflect the reality of what's being done.

Victor-van-Duijnen:
@maxime-poulain Hmm I can let the default behavior for the appsettings.Development if you prefer.

However,  I don't agree with the second statement. Note that the staging environment is reserved more for business people and that you might not want to spam your professionnal individual mail inbox with tests. If Lieven wants to test that feature, as a developer he will override the appsettings.Development to indicate his mail address and check the corresponding behaviour in local or dev env'.
Also adding logic to discriminate smart user from trainer is not worth it and add a unrequired complexity in the code.
That + the fact we can span as many dev environments with specific settings that we want and then I really don't see the point for it.

src/Showcase/src/Smart.FA.Catalog.Showcase.Web/appsettings.Development.json
Victor-van-Duijnen: Hmm that's a business request  @LievenPraet. Apparently product@smart.coop is the closest to a staging address for testing in cfa that we have :/ Maybe @juliettebej or @Geraldineb1 want to intervene on that point.

As usual I am open for naming changes but you have to give me some bones/suggestions there @maxime-poulain @LievenPraet

maxime-poulain:
>  Note that the staging environment is reserved more for business people 

The issue is that the stage is supposed to be a mirror of the production, isn't it ? Hence the PreProduction naming we use.

>  you might not want to spam your professionnal individual mail inbox with tests

That would not be the point, it would be to, in this case, to ensure, to reproduce the actual behavior it happens in production.

> If Lieven wants to test that feature, as a developer 

Well it applies at @juliettebej  or @Geraldineb1 , etc.

Victor-van-Duijnen:
> The issue is that the stage is supposed to be a mirror of the production, isn't it ? Hence the PreProduction naming we use.

It's a mirror of production made for testing. You already agree that - if it's not a smart permanent member- the mail will be sent to another address so there would already be variation of the Production env' anyway(that's the whole point of that PR btw).

>That would not be the point, it would be to, in this case, to ensure, to reproduce the actual behavior it happens in production.

In this case the only logic changing is the GetToRecipient,(that's why it's inherited from the component used in production).

>Well it applies at @juliettebej or @Geraldineb1 , etc.

@juliettebej and @Geraldineb1 have access to product@smart.coop (it's the point of that address, and all the testing business team normally).

maxime-poulain:
I understand your point. 

I just point out that if one wants to test the actual full workflow he just cant (one way or another) unless he/she  does it in production or is; of course, a developer.

For further development it might be better to have settings that switch on/off a functionality instead of hardcoding a behavior depending of the underlying environment the code runs onto. 
This settings default configuration can be then be made accordingly of the environment.

> In this case the only logic changing is the GetToRecipient,(that's why it's inherited from the component used in production).

I know I have made the remark.

But it's a bit of too much talk from my side: like you said it's gonna work just fine the way it is.

Good work !

maxime-poulain (approved):

src/Showcase/src/Smart.FA.Catalog.Showcase.Infrastructure/Mailing/Inquiry/Trainer/TrainerInquirySendTestRecipientEmailService.cs
Victor-van-Duijnen: @LievenPraet What about just `TestInquirySendEmailService` since instead of sending the mail to a trainer's email, it sends it to a a product team email (was thinking also of `ProductTeamInquirySendEmailService `but it's longer)?

PR #259, opened 2022-07-14
Specflow testing with business

Victor-van-Duijnen:
[This PR ](https://smartjira.atlassian.net/browse/CFA-125)intends to be part of the workflow for creating and editing of unit tests for business and developer using Specflow.
iIt only targets defining part of a process for business unit test creation with easy dev integration, NOT be exhaustive with all the tests, yet (other PR).


src/UserAdmin/test/Smart.FA.Catalog.Tests.Specs/Steps/TrainerStepDefinition.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        _action.Should().Throw<Exception>();
+    }
+
+
+    [When(@"I try to update with the invalid title (.*)")]
+    public void WhenITryToUpdateWithTheInvalidTitle(string title)
+    {
+        _action = () => _trainer.UpdateTitle(title);
+    }
```
LievenPraet, line 36: This function is already defined above

LievenPraet (changes requested):

src/UserAdmin/test/Smart.FA.Catalog.Tests.Specs/Steps/TrainerStepDefinition.cs
Victor-van-Duijnen: Well it was a test scenario for adding title as "examples" (Specflow feature) or directly hardcode the title. Following your comment,  I have removed it though for avoiding confusion.

LievenPraet (approved):

src/UserAdmin/test/Smart.FA.Catalog.Tests.Specs/Features/Trainers.feature
```diff
@@ -0,0 +1,12 @@
+Feature: test cases relating to creation and edition of a trainer
+
+    @CFA-188
+    Scenario: Throw exception when trying to create a trainer with an invalid title (length too big)
```
maxime-poulain, line 4:
```suggestion
    Scenario: Throw exception when trying to update a trainer with an invalid title (length too big)
```

Smart.FA.Catalog.sln
```diff
@@ -124,6 +124,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "compose", "compose", "{694A
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Smart.FA.Catalog.Showcase.Infrastructure.Tests", "src\Showcase\test\Smart.FA.Catalog.Showcase.Infrastructure.Tests\Smart.FA.Catalog.Showcase.Infrastructure.Tests.csproj", "{17634CDE-4641-4C03-A32C-135658339BE7}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Smart.FA.Catalog.Tests.Specs", "src\UserAdmin\test\Smart.FA.Catalog.Tests.Specs\Smart.FA.Catalog.Tests.Specs.csproj", "{4C06E7E3-8C07-4CE2-9AB2-E74BE5DF180B}"
```
maxime-poulain, line 127: Naming the project "Specflow" or "SpecFlow" wouldn't be better to reflect the framework's name ?

src/UserAdmin/test/Smart.FA.Catalog.Tests.Specs/Features/Trainers.feature
```diff
@@ -0,0 +1,12 @@
...
+        Given I have a valid trainer
```
maxime-poulain, line 5: Would the following be better ?

Given A user wants to update its profile
When the title he entered is too long
then an error should be thrown ?

maxime-poulain (changes requested):

Victor-van-Duijnen:
**Note to myself for specflow tests:** 

1. Whitespaces between specflow reserved keywords is very important => `Scenario : (...)` instead of `Scenario: (...)` will actually break the whole specflow file (and worse, without notice).
2. When writing multiple times the same genkins sentence (for generic cases like "Then the code should throw an exception"), a scope attribute is necessary to target the proper method in the step definitions BUT the IDE doesn't support dynamic navigation EVEN with the scope attribute.
3. Obvious but point of attention on the number of columns in Examples which absolutely need to match the number of columns defined in the header.

PR #260, opened 2022-07-15
Change prod sql server cn name

Victor-van-Duijnen:
Change of the connectring string database (public website and administrator) for targeting the new sql server in production.


maxime-poulain:
You need also to remove the EF Core workaround for pagination.

PR #265, opened 2022-07-20
Replace redirect while preserving modelstate with simple page return

Victor-van-Duijnen:
Serializing huge modelstate in tempdata can create huge cookies that would span 431 http request


Victor-van-Duijnen:
Serializing huge modelstate in tempdata can create huge cookies that would span 431 http request

PR #271, opened 2022-07-26
[ShowCase] Use resources for search bar

NourJammal:
# ShowCase
Use resources for search bar in training list

https://smartjira.atlassian.net/browse/CFA-544


src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingList/TrainingList.cshtml
```diff
@@ -20,7 +20,7 @@
     <section class="section__training-search">
         <form method="get">
             <div class="c-input-with-icon">
-                <input class="c-input c-input--max-width" id="searchKeyword" name="@nameof(Model.SearchKeyword)" value="@Model.SearchKeyword" type="search" placeholder="Search…">
+                <input class="c-input c-input--max-width" id="searchKeyword" name="@nameof(Model.SearchKeyword)" value="@Model.SearchKeyword" type="search" placeholder="@ShowcaseResources.SearchLabel…">
```
Victor-van-Duijnen, line 23: I might be a nitpicker but I feel the end "..." should be part of the localized ressource. Since it's PR related, and that eventually what comm' will receive are the localized files, they might want to refomulate that differently and remove the "..." in their new formulation.

Victor-van-Duijnen (approved):

PR #276, opened 2022-08-04
[Showcase] Change appareance of given by smart text

joe-coop:
- Change the appareance of GivenBySmart text in Training card
- Add css styles for the bloc text GivenBySmart
- Add new localization keys for the bloc text GivenBySmart


src/Showcase/src/Smart.FA.Catalog.Showcase.Web/Pages/Training/TrainingDetails/TrainingDetails.cshtml
```diff
@@ -38,6 +34,15 @@
             }
         }
     </div>
+    @if (Model.Training.IsGivenBySmart)
```
Victor-van-Duijnen, line 37: I would try to avoid as much as possible if/else logic in our main views. Could we move the whole logic of that if/else statement inside a partial view. The partial view can take a custom model including a text and a condition for displaying or not the givenBySmart logo. I have created a new branch - feature/cfa-551-small-refactor  - based on that one implementing the partial view, feel free to merge it with this branch!

Victor-van-Duijnen (approved):

